PR #4, opened 2022-03-16
Discord experimental

ajCameron:
So far can only send messages back to the channel the original message came from.

Will work on it soem more.


.gitignore
```diff
@@ -5,6 +5,14 @@
 /.local/
 /.cache/
 /venv/
+.idea
```
javajawa: Set up your global gitignore up correctly! :3

src/mewbot/ioconfigs/post_input/input.py
```diff
@@ -0,0 +1,107 @@
...
+        app.add_routes(
+            [
+                web.post("/", self.post_response),
+            ]
+        )
+
+        self.logger.info("About to run web app")
+        # Re. The protected member access
+        # I'm not sure why the aiohttp people have made some of the design choices they have - perhaps this is not
+        # necessary and I've just misunderstood
```
javajawa: https://github.com/javajawa/snerge-bot/blob/aiohttp/src/snerge/__main__.py#L64-L77

src/mewbot/ioconfigs/discord_io/input.py
```diff
@@ -0,0 +1,102 @@
...
+    def produces_inputs() -> Set[Type[InputEvent]]:
+        """
+        Defines the set of input events this Input class can produce.
+        :return:
+        """
+        return {
+            DiscordInputEvent,
+        }
+
+    async def run(self) -> None:
```
javajawa: Does creating a task inside the run make sense? Doing so seems to make it harder for the caller to determine if something has fallen over, as it won't know what are the relevant tasks

ajCameron: There was a reason I did it that way ... but I can't remember for the live of me what it was. So it probably wasn't great. Correcting.

src/mewbot/ioconfigs/post_input/input.py
ajCameron: Implemented.

ajCameron:
Think it's ready when you are - if you want to take a final look?

(Now, mypy is failing locally but passing remotely ... which, you know. Better than the other way round). :)

ajCameron:
Everything even still works ... somehow,

javajawa (dismissed):

design-docs/README.md
```diff
@@ -108,12 +108,14 @@ with the bot.
 
 #### IOConfig
 
-From an implementation perspective, these will be configures as inputs/outputs
+From an implementation perspective, these will be configured as inputs/outputs
 (so cases like discord where one connection is both don't require two to be
 configured). These configurations will then present a list of Inputs and Outputs
 which can be used in Triggers.
 
 ```python
```
exachixkitsune, line 116: Documentation Mismatch with Code : This does not appear to be a duplicate of any code anywhere in mewbot. Where should it be?

javajawa: It's demonstrative rather than actual code. pycharm assumes that python code blocks should be correct and happens to lint them as well, so I fixed them up whilst doing the refactoring.

examples/trivial_discord_bot.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+        self.logger.info(test_event)
+        await self.send(test_event)
+
+    async def send(self, event: OutputEvent) -> None:
+        await self._queue.put(event)
+
+
+def load_secrets() -> Any:
```
exachixkitsune: Future Thoughts: Should this end up being a default/inbuilt thing to mewbot?

src/mewbot/behaviour.py
```diff
@@ -75,29 +21,31 @@ def __init__(self) -> None:
         self.conditions = []
         self.actions = []
 
-    def add(self, block: BehaviourComponent) -> None:
-        if not isinstance(block, BehaviourComponent):
+    def add(
```
exachixkitsune: Future Thoughts: One "setup" function which takes in the three Trigger / Condition / Action
Use case: setting up a behaviour on one line in bot init

src/mewbot/behaviour.py
```diff
@@ -75,29 +21,31 @@ def __init__(self) -> None:
...
 
         self.interests = interests
 
-        if isinstance(block, Trigger):
-            self.triggers.append(block)
-        if isinstance(block, Condition):
-            self.conditions.append(block)
-        if isinstance(block, Action):
-            self.actions.append(block)
+        if isinstance(component, TriggerInterface):
```
exachixkitsune: Is there a potential use case for an induvidual component to be an instance of multiple of these? Can I have a component that is both a TriggerInterface and a ConditionInterface?
Do we want to accept those conditions, warn on those conditions, or terminate?

src/mewbot/behaviour.py
```diff
@@ -2,70 +2,16 @@
...
 
 
 class Behaviour:
     active: bool = True
-    triggers: List[Trigger]  # This of things that can trigger this behaviour
-    conditions: List[
-        Condition
-    ]  # conditions on the action being called when it is matched (e.g. "is subscribed")
-    actions: List[Action]  # the things to do if an event matches, and the conditions are met
+    triggers: List[TriggerInterface]
```
exachixkitsune: Minor/Readibility: Comments on each line were lost

src/mewbot/behaviour.py
javajawa: Not needed, configuration should never be in python

src/mewbot/behaviour.py
javajawa: Doing so is explicity forbidden in the component API (https://github.com/ajCameron/mewbot/blob/2990b63/src/mewbot/component.py#L43-L46) but nothing prevents alternative implementations of these interfaces doubling up.

examples/trivial_discord_bot.py
```diff
@@ -0,0 +1,148 @@
...
+    print_behavior = Behaviour()
+    # Triggers on all PostInput events
+    print_behavior.add(DiscordTrivialTrigger())
+    # Passes all incoming events
+    print_behavior.add(DiscordTrivialCondition())
+    # Prints all the events which make their way into the bot
+    print_behavior.add(DiscordPrintAndReplyAction())
+
+    # - IOConfig
+    post_ioconfig = PostIOConfig()
```
exachixkitsune: Future Thoughts/Shortcuts:If PostIOConfig is required for configuring Discord, do we want to make a system that automatically does that pipeline?
Or do we only ever do one add_io_config to bot, and so post_ioconfig should contain everything?

src/mewbot/behaviour.py
javajawa: Intentional as inline comments should not be used for documentation (probably a hold over from copy pasting from the design documents?)
Currently pretty much everything is missing both doucmentation and tests

examples/trivial_discord_bot.py
javajawa: Neither, really.
Building on both https://github.com/mewler/mewbot/tree/2990b637184105bdba773f0cff50007f968632f7/design-docs#input and the principle of configuration of the bot not being driven by python code, the solution is that the `mewbot.loader` will be able to load IOConfig objects of arbitary classes as needed, and those classes will automatically output the Input/Outputs that they support. (The method of building it up here is essentially the wrong way around)

src/mewbot/http_post.py
```diff
@@ -0,0 +1,117 @@
...
+    to add yourself - and that's about it.
+    """
+
+    inputs: List[Input]
+    outputs: List[Output]
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        self.inputs = []
```
exachixkitsune: Minor/Missing:These have gets but not sets/adds; https://github.com/mewler/mewbot/blob/fe092a331db784ebb88f04028136a919a3f1656c/examples/trivial_discord_bot.py#L141 adds to them.

src/mewbot/behaviour.py
javajawa: (more likely is that this will get changed to adding an iterable of that the current type, a la
```python
add(self, *components: Union[ *Interface... ]
```
and then adding all of them by the existing method.

examples/trivial_discord_bot.py
javajawa: Secret storage in the configuration language (and the YAML backing store) is something I've not really got a solution for yet. Thoughts on that welcome in the design docs.

src/mewbot/bot.py
```diff
@@ -4,28 +4,38 @@
...
 
 class Bot:
     name: str  # The bot's name
-    io: List[IOConfig]  # Connections to bot makes to other services
+    io_configs: List[IOConfigInterface]  # Connections to bot makes to other services
     behaviours: List[Behaviour]  # All the things the bot does
     datastores: Dict[str, DataSource[Any]]  # Data sources and stores for this bot
 
+    def __init__(self) -> None:
+        self.name = "not_set"
```
exachixkitsune: Minor/Missing:This has no inherent gets or sets.

examples/trivial_discord_bot.py
```diff
@@ -0,0 +1,148 @@
+# Remember when wiring random services together to see what you could make happen?
```
exachixkitsune: Overall Q for trivial_discord_bot.py; is this the eventual workflow we're going for with mewbot, or is this a temporary example for dev?

setup.py
```diff
@@ -11,14 +11,14 @@
 setuptools.setup(
     name="mewbot",
     version="0.0.1",
-    author="Benedict Harcout $ Alex Cameron",
-    author_email="[Ben public email for this project] & mewbot@quicksilver.london",
+    author="Benedict Harcourt & Alex Cameron",
+    author_email="mewbot@tea-cats.co.uk & mewbot@quicksilver.london",
     description="Light weight text based generic irc framework",
```
exachixkitsune: Should be "YAML-driven Bot Framework", or something similar?

src/mewbot/demo.py
```diff
@@ -0,0 +1,32 @@
...
+class Foo(Condition):
+    @staticmethod
+    def consumes_inputs() -> Set[Type[InputEvent]]:
+        return set()
+
+    _channel: str
+
+    def __init__(self) -> None:
+        if not hasattr(self, "_channel"):
+            self._channel = "oops"
```
exachixkitsune: I know it's only a demo; should this error or be otherwise a warning; or does this transit to a warning/error later because "oops" is generally not a channel?

src/mewbot/bot.py
exachixkitsune: Apparently unused unless I cannot find where it's used.

src/mewbot/bot.py
```diff
@@ -68,46 +84,75 @@ class BotRunner:
...
+
+        # Seemed to fix a shutdown bug - now less sure
+        asyncio.ensure_future(self.create_tasks(loop), loop=loop)
+        # Need a running event loop for some of the io init tasks
+
+    async def create_tasks(self, loop: asyncio.AbstractEventLoop) -> None:
+        self.logger.info("Starting main event loop")
+
+        # Startup the inputs
+        self.logger.info("self.inputs: %s", ",".join(map(str, self.inputs)))
```
exachixkitsune: Not checked, but want to note a moment of care may need to happen here if the stringification of self.inputs captures the configuration of each input and any auth tokens if they are bound up in that. (which then gets logged presumably...)

src/mewbot/loader.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Type, Any
+
+from mewbot.core import ComponentKind
+from mewbot.component import ConfigBlock, Component, ComponentRegistry
+
+
+def load_component_from_yaml(config: ConfigBlock) -> Component:
+    required_keys = ConfigBlock.__annotations__.keys()  # pylint: disable=no-member
+
+    if set(config.keys()) != set(required_keys):
```
exachixkitsune: Should this be != or should it be A contains B?

exachixkitsune (dismissed):
oops accidently added a bunch of comments before someone let me know that was wrong.
I've resolved any "conversations" which looked like my query had been answered.
No major comments, only minors. And, let's be honest, many of those are possibly me pointing out things that are like that for a reason.

src/mewbot/bot.py
javajawa: Field is included as per spec (https://github.com/mewler/mewbot/tree/mainline/design-docs#bot) but doesn't actually really do anything -- at least until building out the UI and API structures starts happening.

src/mewbot/bot.py
javajawa: By the spec, they should not just be dumping dicts of their own state (https://github.com/mewler/mewbot/tree/mainline/design-docs#human-readable-descriptions), but the burden of defining sensible str representation of components is up to the people implementing that component.

src/mewbot/loader.py
javajawa: Great question!
I left this as a direct equality check because it means that extra incoming properties will cause an error, making schema mistakes easier to detect.

There's a reasonable chance that there will be multiple TypedDicts representing different complex structures (such as a behaviour having triggers/conditions/actions) which is going to make this section of code more complex.

src/mewbot/demo.py
javajawa: This is in general an interesting question because of the order of operations.
With the way the metaclass is setup, `ComponentRegistry.__call__` invokes `Component.__new__`, then sets all settable property decorators, then calls `Component.__init__` with any parameters not previously used. Mostly it is done in this order so that calls to create an object don't rely toooo much on the class's `__init__`, and also that that `__init__` may assume that all the properties are in place.

However, at the same time, because python classes are ducktyped, if a property is not set in the `__new__`, we still need a way to have the `__init__` create the underlying property if it was not set (or error out).

Potential options for that are:
 - Forcing all properties to be required in the `__new__` call (i.e. you can not not specify the properties) and thus ensuring the property is always defined
 - Default value on the property in the class definition, which will get overwritten by any call from `__new__`
 - This kind of checked operation in current code
 - Change how the loader works that setup happens after `__init__`

src/mewbot/loader.py
javajawa: This has been changed to require a superset, as that was necessary for Behaviour loading.

src/mewbot/demo.py
javajawa: I have solved this specific example by just not having an `__init__` function, and a field default of `""`

examples/trivial_discord_bot.py
javajawa: https://github.com/mewler/mewbot/pull/4/commits/c8c08db127b13daa3d56f1c157a72cbbc2468fe4 is a good example of how it is envisioned to work!

src/mewbot/http_post.py
javajawa: I think I have fixed this!

examples/trivial_discord_bot.py
ajCameron: My aim when putting it together was very tempoary example for dev. Just wanted something with a working event loop to play around with/preform basic validation of class drafts.

"example" is a bad name for this concept. In future, will tag such things as "scratch_{name}.py"

examples/trivial_discord_bot.py
ajCameron: Will also consider it! That's definitely a problem that has to be solved.

setup.py
ajCameron: I definitely agree. Will update.

ajCameron:
Leaving in existence incase squash was the wrong way to go

PR #23, opened 2022-04-05
Desktop notification windows

ajCameron:
Tested and working on ubuntu 20.04 LTS, WLS and windows.

Some patching required to make mewbot run happily on windows - in particular, can't install the signal handlers.

setup.py modified to install win10toast if on windows.

### DIFF
--- a/examples/discord_to_desktop_notification.yaml
+++ b/examples/discord_to_desktop_notification.yaml
@@ -1,0 +1,44 @@
+apiVersion: v1
+kind: IOConfig
+module: mewbot.io.discord
+name: DiscordIO
+uuid: 1
+properties:
+  token: "[token goes here]"
+
+---
+
+apiVersion: v1
+kind: IOConfig
+module: mewbot.io.desktop_notification
+name: DesktopNotificationIO
+uuid: 2
+properties:
+  token: "[should-not-be_needed]"
+
+---
+
+apiVersion: v1
+kind: Behaviour
+module: mewbot.api.v1
+name: Behaviour
+uuid: 1
+properties:
+  name: 'Echo Inputs'
+triggers:
+  - apiVersion: v1
+    kind: Trigger
+    module: examples.discord_to_desktop_notification
+    name: DiscordTextCommandTrigger
+    uuid: 1
+    properties:
+      command: "!hello"
+conditions: []
+actions:
+  - apiVersion: v1
+    kind: Action
+    module: examples.discord_to_desktop_notification
+    name: DiscordMessageToNotificationAction
+    uuid: 1
+    properties:
+      message: "world"
--- a/examples/discord_to_desktop_notification.yaml
+++ b/examples/discord_to_desktop_notification.yaml
@@ -13,8 +13,7 @@ kind: IOConfig
 module: mewbot.io.desktop_notification
 name: DesktopNotificationIO
 uuid: 2
-properties:
-  token: "[should-not-be_needed]"
+properties: {}
 
 ---
 
--- a/examples/new 3.txt
+++ b/examples/new 3.txt
@@ -1,0 +1,2 @@
+    # ['summary_detail', 'published_parsed', 'links', 'title',
+    #  'summary', 'guidislink', 'title_detail', 'link', 'published', 'id']
--- a/examples/new 3.txt
+++ b/examples/new 3.txt
@@ -1,2 +1,0 @@
-    # ['summary_detail', 'published_parsed', 'links', 'title',
-    #  'summary', 'guidislink', 'title_detail', 'link', 'published', 'id']
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,10 @@ setuptools.setup(
         f"License :: OSI Approved :: {true_licence}",
         "Operating System :: OS Independent",
     ],
+    extras_require={
+        ':sys_platform == "win32"': [
+            'win10toast~=0.9'
+        ], },
     package_dir={"": "src"},
     packages=setuptools.find_packages(where="src"),
     python_requires=">=3.8",  # Might be relaxed later
--- a/setup.py
+++ b/setup.py
@@ -26,9 +26,8 @@ setuptools.setup(
         "Operating System :: OS Independent",
     ],
     extras_require={
-        ':sys_platform == "win32"': [
-            'win10toast~=0.9'
-        ], },
+        ':sys_platform == "win32"': ["win10toast~=0.9"],
+    },
     package_dir={"": "src"},
     packages=setuptools.find_packages(where="src"),
     python_requires=">=3.8",  # Might be relaxed later
--- a/setup.py
+++ b/setup.py
@@ -25,9 +25,6 @@ setuptools.setup(
         f"License :: OSI Approved :: {true_licence}",
         "Operating System :: OS Independent",
     ],
-    extras_require={
-        ':sys_platform == "win32"': ["win10toast~=0.9"],
-    },
     package_dir={"": "src"},
     packages=setuptools.find_packages(where="src"),
     python_requires=">=3.8",  # Might be relaxed later
--- a/src/mewbot/io/desktop_notification.py
+++ b/src/mewbot/io/desktop_notification.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Optional, Set, Sequence, Type, Callable
+from typing import Optional, Set, Sequence, Type
 
 import dataclasses
 import logging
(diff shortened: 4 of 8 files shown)


ajCameron:
Something seems to be going really badly wrong with loading more than one IOConfig.

Like - I got it working, but please note 

apiVersion: v1
kind: IOConfig
module: mewbot.io.desktop_notification
name: DesktopNotificationIO
uuid: 2
properties:
  token: "[should-not-be_needed]"

examples/discord_to_desktop_notification.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+---
+
+apiVersion: v1
+kind: IOConfig
+module: mewbot.io.desktop_notification
+name: DesktopNotificationIO
+uuid: 2
+properties:
+  token: "[should-not-be_needed]"
```
javajawa: You might be able to make the properties problem go away with
```suggestion
properties: {}
```
(unless you have already tried that).

src/mewbot/io/desktop_notification.py
```diff
@@ -0,0 +1,242 @@
...
+    def __init__(self) -> None:
+
+        self._logger = logging.getLogger(__name__ + "DesktopNotificationOutputEngine")
+        self._logger.info("DesktopNotificationOutputEngine starting")
+
+        # Note-to-self: Warn if we are in a non-tested state
+        # Note-to-self: Need to test to see if we are headless and fallback if we are
+
+        self._platform_str = sys.platform
+        self._logger.info(f"We are detected as running on {self._platform_str}")
```
javajawa: I'm surprised that nothing is complaining about this not being lazily evaluated
```suggestion
        self._logger.info("We are detected as running on %s", self._platform_str)
```

(I'm not going to repeat the comment for each instance, but I should go look at why the linters aren't picking that up)

src/mewbot/io/desktop_notification.py
```diff
@@ -0,0 +1,242 @@
...
+        # is required.
+        # silent failure of desktop notifications is not desired. Noisy failure is.
+
+        self._enabled = False
+        if self._platform_str == "win32" or self._platform_str == "win64":
+            self._detected_os = "windows"
+            self._do_windows_setup()
+
+        elif self._platform_str == "linux":
+            if self._platform_str == "linux2":
```
javajawa: This is an impossible condition? (it can not be equal to both...surely)

src/mewbot/io/desktop_notification.py
```diff
@@ -0,0 +1,242 @@
...
+                    f"You seem to be running python below 3.3, "
+                    f"or python behavior has changed - sys.platform = {self._platform_str}"
+                )
+
+            self._detected_os = "linux"
+            setattr(self, "_notify", self._linux_notify_send_method)
+
+        elif self._platform_str == "darwin":
+            self._detected_os = "macos"
+            self._enabled = False
```
javajawa, line 138: No logging information about the OS in this branch.
Might be worth pulling out to a single log statement at the bottom with the detected OS and the enabled state?

src/mewbot/io/desktop_notification.py
```diff
@@ -0,0 +1,242 @@
...
+        if not isinstance(event, DesktopNotificationOutputEvent):
+            return False
+
+        return self._engine.notify(event.title, event.text)
+
+
+class DesktopNotificationOutputEngine:
+    """
+    Contains the dubious and concerning logic required to make output work on all target systems.
+    """
```
javajawa: The doc block should at least mention what this is and how it works ;)

```suggestion
    """
    Cross-Platform notification shim/engine.
    
    Works by patching the `_notify` function at init time to contain the correct implementation.
    """
```

src/mewbot/io/desktop_notification.py
```diff
@@ -0,0 +1,242 @@
...
+        if not self._enabled:
+            return False
+        return self._notify(title, text)
+
+    def _notify(self, title: str, text: str) -> bool:
+        self._logger.warning(
+            f"{self._notify} should never be called directly - "
+            f"should have been live replaced with the true method - f{title} - f{text}"
+        )
+        return False
```
javajawa: An alternative to this would be a dict of [detected OS; function to call] which might be easier to follow:
```suggestion
    def _notify(self, title: str, text: str) -> bool:
        # Probably don't do this inline...
        caller = {"windows": self._windows_toast_method, "linux": self._linux_notify_send_method}.get(self._detected_os)
        
        if not caller:
          self._logger.warning("No notification method supported for %s", self._detected_os)
          return False
          
        return caller(title, text)
```

javajawa (commented):
Looks good codewise -- I've left a few comments here and there but nothing major.
Would probably be good to rewrite some of the commit messages to be more descriptive of the changes

setup.py
```diff
@@ -25,6 +25,9 @@
         f"License :: OSI Approved :: {true_licence}",
         "Operating System :: OS Independent",
     ],
+    extras_require={
+        ':sys_platform == "win32"': ["win10toast~=0.9"],
```
exachixkitsune: Should this be added to requirements.txt?

src/mewbot/io/desktop_notification.py
```diff
@@ -0,0 +1,242 @@
...
+        self._logger.info(f"We are detected as running on {self._platform_str}")
+
+        # Could use "in" or string normalisation - but I want it to fail noisily and log the error
+        # if I have made any incorrect assumptions.
+        # So, hopefully, the user will tell us and I can check to see if anything creative
+        # is required.
+        # silent failure of desktop notifications is not desired. Noisy failure is.
+
+        self._enabled = False
+        if self._platform_str == "win32" or self._platform_str == "win64":
```
exachixkitsune: win64 doesn't appear to ever be a valid output from sys.platform
(source: https://docs.python.org/3.10/library/sys.html#sys.platform , not reporting win64 on my version)

src/mewbot/io/desktop_notification.py
exachixkitsune: Does the setattr method appreciably reduce runtime?

src/mewbot/io/desktop_notification.py
exachixkitsune: How tight is the setup.py's requirement on python 3.8+, will this likely ever occur, or is it a catch for edge cases?

exachixkitsune (commented):
Had a look, all seems reasonable, though I've not actually tested the code.

src/mewbot/io/desktop_notification.py
javajawa: `setattr` will be faster than what I've outlined here (as there won't essentially be a proxy to another method) and is almost certainly the fastest method for live matching something in standard python. On the flip side, if there's enough notifications that an extra layer of stack indirection makes a difference to real world performance, there's going to be so many notifications

examples/discord_to_desktop_notification.yaml
ajCameron: I would have been prepared to swear I did ... but it would seem not!

setup.py
ajCameron: Did not realise that you could now - have corrected

src/mewbot/io/desktop_notification.py
ajCameron: Corrected. Also, after looking into it, not sure why the linters are not getting it.

src/mewbot/io/desktop_notification.py
ajCameron: Removing.

src/mewbot/io/desktop_notification.py
ajCameron: Derp! Indeed an impossible condition. Corrected.

It's highly unlikely to ever actually occur - indeed a catch for an edge case.

src/mewbot/io/desktop_notification.py
ajCameron: It is easier to read - correcting.

src/mewbot/io/desktop_notification.py
ajCameron: (What happens when you go off memory from a decade ago - instead of just reading the docs)

src/mewbot/io/desktop_notification.py
ajCameron: ... I mean - that might be helpful...

src/mewbot/io/desktop_notification.py
ajCameron: Agreed and done.

examples/new 3.txt
```diff
@@ -0,0 +1,2 @@
+    # ['summary_detail', 'published_parsed', 'links', 'title',
+    #  'summary', 'guidislink', 'title_detail', 'link', 'published', 'id']
```
javajawa: Scratch file? :3

javajawa (commented):
Code approved. What looks like a scratch file got thrown in, should be easy enough to rebase out ;)

setup.py
javajawa: I also had no idea you could do that. That's...going to be useful elsewhere

PR #31, opened 2022-04-12
add a change

javajawa:

PR #32, opened 2022-04-12
Fix Linters

### LINKED ISSUE #6: Improve CI/Actions
Currently, the CI system has a number of issues:

- [x] The linting of examples is broken
- [ ] If a PR is opened from within the repo, we are still using `pull_request_target`
- [ ] There is no information about how to handle changes to the workflows

javajawa:
The linting toolchain for both local development and GitHub actions has been converted into a python script.
This script is capable of outputting GitHub annotations, which can be used across forks (see #31)

This fixes #6


ajCameron:
If we chuck an assert in before _level is assigned to level ("assert _level in ["error", "warning", "notice"]") this should be enough to clue mypy in as to what the variable should be. This is semantically identicle to the checking already done, but I think mypy may be consfused about that check - perhaps it can't tell if something is modifying the set before the check? Converting it to a frozenset may fix? But it could always be overwritten.

So you'd have something like

_level = _level if _level in LEVELS else "error"
assert _level in ["error", "warning", "notice"]
level: Level = _level

But it ...  lacks in elegance.

Overall the new tooling is looking good though! Just this issue, and then, hopefully, linting is solved.

exachixkitsune (dismissed):
Not something I've done before, so I trust it all works :3

Issue #11, opened 2022-03-30
Add testing harness


javajawa:
In order to have high quality code, we should start unit and integration testing.

 - [x] Select a python testing system and coverage library
 - [x] Create a place for tests to live
 - [x] Set it up to have similar linting as the main sources
 - [x] Set up a test system which supports coverage reporting
 - [x] Integrate it with a `tools/test.sh` and the github actions

exachixkitsune:
What flavour of coverage are you looking for?

javajawa:
Probably statement based rather than line based?

exachixkitsune:
Had a look at pytest and pytest-cov, and I have a MWE; though I need to work out how to run a test that loads a file in...
or a way of doing that omitting the file...

(Further work is looking at something that isn't pytest)

exachixkitsune:
So, pytest and pytest-cov gives us:

- relatively standard testing framework
- Easy integration of the coverage tool (pytest-cov also using the default coverage package, and the current settings we have in `pyproject.toml`.
- Output of pytest-cov into html, xml, or plaintext/annotated source (the html one is nice and useful)
- pytest-cov lists what's missing (line no for line coverage at least)

Downside; it looks like it's merging the statement coverage and branch coverage stats together and the threshold is both combined; and it's an overall percentage.
 I also cannot seem to make it fail if any file is untested.

I don't know how this would integrate into a github action, but I can create it to output stuff and it will hopefully operate properly.

If that's all ok, I will proceed.

javajawa:
That all seems fine.

For the workflow/actions side, you can crib from the same source as I got the rest of the workflow stuff:
https://github.com/javajawa/py-tiny-orm/blob/2ce10aa/.github/workflows/lint.yml#L95-L172. That runs the tests against version python versions on two operating systems, and then combines the coverage reports.

exachixkitsune:
Things to do

- [x] Create initial suite of unit tests
- [x] Create initial suite of functional tests
- [x] Connect into Github Actions/Workflow

PR #34, opened 2022-04-14
Setup Sonar with badges

### LINKED ISSUE #33: Setup Sonar integration
[Sonar](https://www.sonarqube.org/) is an excellent 3rd party code quality tool. We should integrate it with the Github Actions to provide additional insight into code quality.

javajawa:
This is partial work on #33. Full Sonar setup will follow #11 to include coverage and quality gates.

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,3 @@
+sonar.python.version = 3.8, 3.9, 3.10, 3.11
+sonar.coverage.exclusions = examples/**/*
+sonar.cpd.exclusions = examples/**/*
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -4,7 +4,14 @@
 
 name: Linting
 
-on: [pull_request]
+on:
+  push:
+    branches:
+      - mainline
+  pull_request:
+  release:
+    types:
+      - published
 
 jobs:
   lint:
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@ SPDX-License-Identifier: BSD-2-Clause
 
 # MewBot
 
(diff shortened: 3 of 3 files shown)


ajCameron:
https://sonarcloud.io/summary/overall?id=mewler_mewbot

Shows 26 code smells - but the sonarcloud bot is showing 0 -  not sure if this is a _them_ problem or an _us_ problem.

Looks good on the summary tho - so I think analysis is occuring successfully.

javajawa:
For PRs, it only shows issues introduced in the changes; so if something gets approved as "good enough for now", it doesn't complain on every following change.
(It also supports assigning things and leaving comments in Sonar -- freely it's meant to be used as a complete tool chain in and off itself, but mostly I'm content to leave it complaining about security problems)

Kitsune raised a good point elsewhere of why the current listing isn't picking up those 26 issues, and the answer is it is, there's just all suppressed in the TOML file

.sonarcloud.properties
```diff
@@ -0,0 +1,3 @@
+sonar.python.version = 3.8, 3.9, 3.10, 3.11
```
exachixkitsune, line 1: Why 3.8 in here?

exachixkitsune (approved):

.sonarcloud.properties
javajawa: Matches https://github.com/mewler/mewbot/blob/mainline/setup.py#L30
I think 3.8 is what is specified everywhere as the requirement, although the linters do currently run (only) on 3.9

ajCameron:
Cool! Then all good by me.

ajCameron (commented):
Seems a sensible addition and the bot not reporting can be fixed later.

PR #37, opened 2022-04-18
Simplify imports by reexporting core in api

### LINKED ISSUE #13: Split codebase into packages
Currently our setup logic is aiming to ship one `mewbot` package.
As we are aiming for a modular system with lots of extensibility, a monolith package is suboptimal.

Current suggestion:
  - `mewbot-core` contains the base packages `mewbot.core`
  - `mewbot-api` contains the `mewbot.component` and `mewbot.api` packages, and depend on `mewbot-core`. This is the package that will be a dependency of any "third party" modules
  - `mewbot-ui` will contain the GUI website. For now I think this will just contain the YAML loader
  - `mewbot` contains the bot runner itself, and depends on everything else. This will be the package people install when they want to run a bot

javajawa:
This merge performs a number of structural refactors:

By exporting the `InputEvent` and `OutputEvent` base classes in `mewbot.api.v1`, the IO libraries no longer need to import from both that and the `mewbot.core` module. This makes the import diagram simpler, and highlights the difference between the core and the plug and play API.

The restrictions on API versions have been relaxed or removed, with tools like `mewbot.loader` testing against the underlying interfaces to confirm API compatability.


src/mewbot/core.py
```diff
@@ -151,8 +118,51 @@ async def process(self, event: InputEvent) -> None:
...
+            cls.IO_CONFIG: IOConfigInterface,
+        }
+
+        if value in _map:
+            return _map[value]
+
+        raise ValueError(f"Invalid value {value}")
+
+
 __all__ = [
```
exachixkitsune, line 163: Should these have a particular ordering, i.e. same as in file for findability?
 I cannot fathom the suggestions interface so you'll have to give me a bit for a suggestion.

exachixkitsune (dismissed):
I don't see much in the way of issues here, only a convention thingamabob on __all_- but that's a minor thought

src/mewbot/core.py
javajawa: I'm not sure it matters, really? The order in the file is to prevent forward definition issues, the order here was mostly grouped by purpose.
There's no guarantee that the things being exported are even first defined in this file (part of this PR is in fact exporting things from another module)

src/mewbot/loader.py
```diff
@@ -103,11 +96,25 @@ def load_component(config: ConfigBlock) -> Component:
 
     target_class: Type[Any] = getattr(module, config["name"])
 
-    if not issubclass(target_class, Component):
+    # Verify that the implementation class matches the interface we got from
+    # the `kind:` hint.
+    if not issubclass(target_class, interface):
         raise Exception("Trying to load a non-registered class from config")
```
ajCameron: This will be a theme with me when it comes to error messages. I'm a fan of maximal detail to give the user a debugging headstart. Though this does result in messier code.

Suggestion: Include a list of the currently registered classes - ideally broken down by type - so the user can easily see if they've just typod or if there is a more fundamental problem and the object they're trying to specify hasn't been loaded at all.

src/mewbot/loader.py
```diff
@@ -103,11 +96,25 @@ def load_component(config: ConfigBlock) -> Component:
...
 
     component = target_class.__call__(uid=config["uuid"], **config["properties"])
 
-    assert isinstance(component, Component)
+    # Verify the instance implements a valid interface.
```
ajCameron, line 124: Suggest some kind of error message here for if the assert fails.

src/mewbot/core.py
ajCameron: I think, in the general case, re-ordering to increase clarity is a good thing. Not sure how to do it here though. But we should be mindful of that in future - I've seen people draw deep - largely wrong - inferences from structure.

exachixkitsune: I think it could just be left as developer preference and is far and away not a thing we should worry about. I'd put it in the level of minor readability.

src/mewbot/loader.py
javajawa: I'm still not over the fact that python's assert doesn't have `assert([bool], [str])`

ajCameron: ...

That would be really kinda neat.

I mean, they'll probably reject it, but I'm weirdly game for drafting a PEP

src/mewbot/core.py
ajCameron: Makes sense! We can always change it later on a readability pass.

src/mewbot/loader.py
javajawa: It turns out, it does work like that! It's just not obvious because `assert` _doesn't use brackets_.

```python
>>> assert(False, "Hello")
>>> assert False, "Hello"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError: Hello
```

The first one gets interpreted as `Tuple[Bool=False, Str="Hello"]` which is truthy.

(Thanks, I hate it. Also I thought they would have fixed that the same time as they fixed `print`)

ajCameron: Ahhhhhhhhh!

Yes. I, too, am far from a fan!

ajCameron (approved):
Sensible organisation and structural changes!

(Just did another pass through, and nothing jumps out to suggest - all looks good).

PR #38, opened 2022-04-18
YAML Schema

### LINKED ISSUE #8: Add JSONSchema for the YAML definitions
- Remove the redudant fields from the YAML schema
- Create a formalised JSONSchema document which describes our schema

javajawa:
*This change follows #37*

The schema for YAML documents has been simplified to have fewer fields, with the class to load being reduced to `implementation:`, and the `apiVersion` field removed.

As part of this change, some parts of the API were found to no longer require knowledge of each other, and the `mewbot.component` module has been split into `mewbot.config`, `mewbot.api.registry` and `mewbot.api.v1` to better reflect 'ownership' of the different concepts.

Fixes #8


exachixkitsune (commented):
I don't know enough of what's going on to review this in any valuable capacity.

examples/discord_to_desktop_notification.yaml
```diff
@@ -1,43 +1,33 @@
...
-module: mewbot.io.desktop_notification
-name: DesktopNotificationIO
-uuid: 2
+implementation: mewbot.io.desktop_notification.DesktopNotificationIO
+uuid: df56b38a-8499-4cec-95c1-b54192b4328a
 properties: {}
 
 ---
 
-apiVersion: v1
```
ajCameron, line 20: I approve of this simplification. I agree it should never actually matter - unless, for some inexplicable reason - we want two different modules with the same name which differ only in api version. I strongly suggest we never do this.

examples/discord_to_desktop_notification.yaml
```diff
@@ -1,43 +1,33 @@
...
 kind: IOConfig
-module: mewbot.io.desktop_notification
-name: DesktopNotificationIO
-uuid: 2
+implementation: mewbot.io.desktop_notification.DesktopNotificationIO
+uuid: df56b38a-8499-4cec-95c1-b54192b4328a
```
ajCameron: I am not entirely clear as to why we're switching to a real, hard coded uuid. The advantage of the bare numbers was that they where clearer as a placeholder.

Suggest - if possible - something like replacing the uuid with "placeholder-change-me" - or something obviously artificial.

Not entirely sure why the uuids are the same for all components? This seems undesirable. Unless I'm missing a trick somewhere.

examples/trivial_discord_bot.yaml
```diff
@@ -1,34 +1,26 @@
-apiVersion: v1
 kind: IOConfig
-module: mewbot.io.discord
-name: DiscordIO
-uuid: 1
+implementation: mewbot.io.discord.DiscordIO
```
ajCameron, line 2: Loads cleaner. Like it a lot.

src/mewbot/api/v1.py
```diff
@@ -16,10 +26,51 @@
...
+        return output
+
+    @property
+    def uuid(self) -> str:
+        return self._id
+
+    @uuid.setter
+    def uuid(self, _id: str) -> None:
+        if hasattr(self, "_id"):
+            raise Exception("Can not set the ID of a component outside of creation")
```
ajCameron: I approve of this error message!

src/mewbot/loader.py
```diff
@@ -72,35 +78,30 @@ def load_behaviour(config: BehaviourConfigBlock) -> BehaviourInterface:
...
-        importlib.import_module(config["module"])
-
-    if not ComponentKind.has_value(config["kind"]):
-        raise ValueError(f"Invalid component kind {config['kind']}")
-
-    kind = ComponentKind(config["kind"])
-    interface = ComponentKind.interface(kind)
-
-    module = sys.modules[config["module"]]
+    # Identify the kind of component we should be loading, and the interface that implies.
```
ajCameron, line 103: Indeed - absent very, very weird edge cases that we can probably safely ignore, this inference should always be safe.

examples/discord_to_desktop_notification.yaml
ajCameron: Ah! I had missed a trick, and you have a uuid verifier regex in the schema.

Would still recommend using different uuids for each component - to prevent (slightly) confusing examples. My worry is that people copying them would infer you _need_ to have the same uuid for each component in a config. Which, I believe, is not only not true but may lead to some internal confusion.

javajawa: Yes, I probably should -- at some point, we're going to start using these UUIDs for things, and I imagine that will break all the things.
I might change them to `aaaaaaaa-aaaa-4aaa-aaaa-00000000000[0-9]`?

ajCameron: That sounds good to me - hard to mistake that for meaningful structure!

ajCameron (approved):
Did a final run through, and couldn't see anything to suggest changing. All looks very good.

PR #39, opened 2022-04-18
rss experiment input complete

ajCameron:
Output to follow - but the input side is currently functional so seemed a good idea to get it merged in and check the linters are still passing e.t.c.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,6 @@ py-cord~=2.1.3
 
 # Notifications module (windows)
 win10toast~=0.9; sys_platform == "win32"
+
+# RSS reader
+feedparser~=6.0.8
--- a/src/mewbot/io/rss.py
+++ b/src/mewbot/io/rss.py
@@ -2,7 +2,18 @@
 
 from __future__ import annotations
 
-from typing import Optional, Set, Sequence, Type, Any, Union, List, SupportsInt, Iterable
+from typing import (
+    Optional,
+    Set,
+    Sequence,
+    Type,
+    Any,
+    Union,
+    List,
+    SupportsInt,
+    Iterable,
+    Mapping,
+)
 
 import asyncio
 
@@ -143,7 +154,21 @@ class RSSInputState:
     _sites: List[str]  # A list of sites to poll for RSS update events
     _sites_iter: Iterable[str]
     _sites_started: Set[str]  # sites which have undergone startup
-    sent_entries: Set[str]  # Entries which have been put on the wire
+    _sent_entries: Mapping[str, Set[str]]  # Entries which have been put on the wire
+
+    def start(self):
+        """
+        Caculates all internal states based off _sites
+        """
+        self._sites_iter = cycle(iter(self._sites))
+
+        new_sent_entries = {}
+        for new_site in self._sites:
+            if new_site in self._sent_entries.keys():
+                new_sent_entries[new_site] = self._sent_entries[new_site]
+            else:
+                new_sent_entries[new_site] = set()
+        self._sent_entries = new_sent_entries
 
     @property
     def sites(self) -> List[str]:
@@ -152,7 +177,7 @@ class RSSInputState:
     @sites.setter
     def sites(self, new_sites: List[str]) -> None:
         self._sites = new_sites
-        self._sites_iter = cycle(iter(new_sites))
+        self.start()
 
     @property
     def sites_iter(self) -> Iterable[str]:
@@ -167,6 +192,21 @@ class RSSInputState:
         Record that a site has been successfully started.
         """
         self._sites_started.add(site_started)
+
+    def note_event_transmitted(self, site_url: str, site_uid: str) -> None:
+        """
+        Record that an entry with a uid has been put on the wire.
+        Now broken down by site so that we can more easily purge the old entries.
+        When they have been superceeded by new ones.
+        Otherwise they would just accumulate endlessly - a nasty memory leak.
+        """
+        self._sent_entries[site_url].add(site_uid)
+
+    def check_for_event(self, site_url: str, site_uid: str) -> bool:
+        """
+        Check to see if an event is recorded as having been sent for the specified site.
+        """
+        return site_uid in self._sent_entries[site_url]
 
 
 class RSSInput(Input):
@@ -194,8 +234,9 @@ class RSSInput(Input):
             _sites=sites,
             _sites_iter=cycle(iter(sites)),
             _sites_started=set(),
-            sent_entries=set(),
+            _sent_entries={},
         )
+        self.state.start()
 
         # Generare a warning if the polling interval is less than this
         self.short_warning_interval = 60
@@ -356,7 +397,7 @@ class RSSInput(Input):
         for _ in range(0, self.startup_queue_depth):
 
             entry_internal_id = self._get_entry_uid(entry=site_entries[_], site_url=site_url)
-            if entry_internal_id in self.state.sent_entries:
+            if self.state.check_for_event(site_url=site_url, site_uid=entry_internal_id):
                 self._logger.warning(
                     "entry_internal_id - %s - for entry - %s - from site - %s - "
                     "has been seen before!",
@@ -372,7 +413,7 @@ class RSSInput(Input):
                 # Not enough entries in the retrieved feed to exhaust startup requirements?
                 break
 
-            self.state.sent_entries.add(entry_internal_id)
+            self.state.note_event_transmitted(site_url=site_url, site_uid=entry_internal_id)
 
         self.state.note_site_started(site_url)
 
@@ -390,12 +431,13 @@ class RSSInput(Input):
 
             entry_uid = self._get_entry_uid(entry, site_url)
 
-            if entry_uid in self.state.sent_entries:
+            if self.state.check_for_event(site_url=site_url, site_uid=entry_uid):
                 break
 
             await self._send_entry(entry=entry, site_url=site_url, startup=False)
             transmitted_count += 1
-            self.state.sent_entries.add(self._get_entry_uid(entry=entry, site_url=site_url))
+
+            self.state.note_event_transmitted(site_url=site_url, site_uid=entry_uid)
 
         self._logger.info("%s entries from %s transmitted", transmitted_count, site_url)
 
--- a/src/mewbot/io/rss.py
+++ b/src/mewbot/io/rss.py
@@ -483,18 +483,55 @@ class RSSInputEventFactory:
         # This combination is used to determine if events have already been sent
 
         input_rss_event = RSSInputEvent(
-            title=self.extract_title(entry=entry, problem_list=problem_list),
-            link=self.extract_link(entry=entry, problem_list=problem_list),
-            description=self.extract_description(entry=entry, problem_list=problem_list),
-            author=self.extract_author(entry=entry, problem_list=problem_list),
-            category=self.extract_category(entry=entry, problem_list=problem_list),
-            comments=self.extract_comments(entry=entry, problem_list=problem_list),
-            enclosure=self.extract_enclosure(entry=entry),
+            title=self.extract_field(
+                entry=entry,
+                attr_name="title",
+                default=TITLE_NOT_SET_STR,
+                problem_list=problem_list,
+            ),
+            link=self.extract_field(
+                entry=entry,
+                attr_name="link",
+                default=LINK_NOT_SET_STR,
+                problem_list=problem_list,
+            ),
+            description=self.extract_field(
+                entry=entry,
+                attr_name="description",
+                default=DESCRIPTION_NOT_SET_STR,
+                problem_list=problem_list,
+            ),
+            author=self.extract_field(
+                entry=entry,
+                attr_name="author",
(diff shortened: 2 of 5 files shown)


src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+import asyncio
+from typing import Optional, Set, Sequence, Type, Any, Union, List, SupportsInt
+
```
javajawa: This is a personal preference thing, but I like to keep typing separate from the other imports

```suggestion
from typing import Optional, Set, Sequence, Type, Any, Union, List, SupportsInt

import asyncio
```

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+    _input: Optional[RSSInput] = None
+    _output: None
+
+    _sites: List[str]  # A list of sites to poll for RSS update events
+    _polling_every: int  # Total re-poll time for the sites in the _sites list
+
+    def __init__(
+        self,
+        *args: Optional[Any],
+        polling_every: Union[str, bytes, SupportsInt],
```
javajawa: Couldn't this be a `@property`?
(The long term goal is to infer setup information form the property objects, as reflecting on the arguments to `__init__` is tricky)

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+
+    _sites: List[str]  # A list of sites to poll for RSS update events
+    _polling_every: int  # How often to poll all the given sites (in seconds)
+    sites_started: Set[str]  # sites which have undergone startup
+    sent_entries: Set[str]  # entries which have been put on the wire
+
+    def __init__(
+        self, sites: List[str], polling_every: int, startup_queue_depth: int = 5
+    ) -> None:
+        super(Input, self).__init__()  # pylint: disable=bad-super-call
```
javajawa:
```suggestion
        super().__init__()
```
newer python is smarter for single-inheritance; the pylint warning is poorly worded in explaining what it wants you to do

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+                asyncio.get_running_loop().create_task(self.startup_site_feed(target_site))
+                # Delay before getting the next site
+                # stagger site reads to prevent overwealming anything
+                await asyncio.sleep(self.polling_every)
+                continue
+
+            asyncio.get_running_loop().create_task(self.poll_site_feed(target_site))
+            # Delay before getting the next site
+            # stagger site reads to prevent overwealming anything
+            await asyncio.sleep(self.polling_every)
```
javajawa: I'm not sure if it improves readability, but maybe

```suggestion
        for target_site in self._sites_iter:

            if target_site not in self.sites_started:
                future = self.startup_site_feed(target_site)
            else:
                future = self.poll_site_feed(target_site)

            asyncio.get_running_loop().create_task(future)
            # Delay before getting the next site
            # stagger site reads to prevent overwealming anything
            await asyncio.sleep(self.polling_every)
```

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+            self.sent_entries.add(entry_internal_id)
+
+        self.sites_started.add(site_url)
+
+    async def poll_site_feed(self, site_url: str) -> None:
+
+        self._logger.info("Reading from site %s", site_url)
+
+        # Read the site feed
+        site_newsfeed = feedparser.parse(site_url)
```
javajawa: Is feedparser using asyncio? If not, can we do the HTTP connection with aiohttp and then give it the content, so there isn't a blocking call here?

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+        # Some RSS entries do not confirm to the spec - so some creative parsing might be required
+        # Starting basic, and building up later
+
+        problem_list: List[str] = []
+
+        # guid - there are no rules for this syntax
+        # THUS ONLY THE COMBINATION OF GUID AND SITE_URL SHOULD BE ASSUMED UNIQUE
+        # This combination is used to determine if events have already been sent
+
+        input_rss_event = RSSInputEvent(
```
javajawa: Potential improvement, again don't know if it would actually improve readability: take this and all the functions called creating this object, and put them in a Builder/Marshler object -- essentially this class deals with wire transit (or Extract/Load parts of ETL) and have the transformation be distinct code.

(Also means you could possibly do weird plug-and-play stuff with having different marshlers in the future)

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+        try:
+            entry_guid = entry.guid
+        except AttributeError:
+            problem_list.append(
+                f"entry did not possess a guid - "
+                f"{site_url} might be producing malformed RSS entries"
+            )
+
+            # This is not a fantastic guid - but one needs to be set for internal tracking
+            entry_guid = f"{site_url}:{entry_title}"
```
javajawa: Minor: I'd be more included to go with link and date

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+            # No logging as this is an optional attribute
+            entry_enclosure = ENCLOSURE_NOT_SET_STR
+
+        return str(entry_enclosure)
+
+    def _extract_guid(
+        self, entry: feedparser.util.FeedParserDict, site_url: str, problem_list: List[str]
+    ) -> str:
+
+        entry_title = self._extract_title(entry=entry, problem_list=problem_list)
```
javajawa: This is only used in the `AttributeError` block, possible doesn't want to be up here

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,514 @@
...
+                break
+
+            await self._send_entry(entry=entry, site_url=site_url, startup=False)
+            transmitted_count += 1
+            self.sent_entries.add(self._get_entry_uid(entry=entry, site_url=site_url))
+
+        self._logger.info("%s entries from %s transmitted", transmitted_count, site_url)
+
+    @staticmethod
+    def _extract_title(entry: feedparser.util.FeedParserDict, problem_list: List[str]) -> str:
```
javajawa: These field extractors appear to have a lot of common code. Would something similar to the `Dict.get()` semantics be easier to work with, where the user supplies the default in? The additions to `problem_list` could be based off a list of required fields.

Alterantively
```python
REQUIRED_FIELDS = {
  "title": "TITLE_NOT_SET",
  ...
}

def _extract_field(entry, field: str, problem_list: List[str]) -> Optional[str]:
  if has_attr(entry, field):
    return get_attr(entry. field)
  if field not in REUQIRED_FIELDS:
    return None
    
  problem_list.append(f"entry did not contain {field}")
  return REQUIRED_FIELDS[field]
```

javajawa (commented):
Code looks solid; got a few questions and ideas for simplifying things.

I'd like to sequence this after #37 and #38 which changes some of the underlying APIs; I think the best thing to do will be to rebase this branch once those are merged to keep a linear history, and also remove some of the linting commits that have no semantic value.

src/mewbot/io/rss.py
javajawa: `asyncio.get_running_loop()` could even be determined at `__init__` time, to be honest

exachixkitsune: I meant to say "I agree with this, it removes an implicit branch and also avoids a weird branch to look at in the code" but SOMEONE misread "Commit Suggestion" as "Comment on suggestion"
(it was me, I misread github)

requirements.txt
```diff
@@ -8,4 +8,6 @@ py-cord~=1.7
 aiohttp~=3.7.4  # Waiting for py-cord to upgrade
 setuptools~=61.1
 
-win10toast~=0.9; sys_platform == "win32"
\ No newline at end of file
+win10toast~=0.9; sys_platform == "win32"
+
+feedparser~=6.0.8
```
exachixkitsune:
```suggestion
feedparser~=6.0.8

```
Just adding a newline here in requirements.txt to avoid it thinking something changed with the prev line when nothing changed.

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,512 @@
...
+    source: str  # The RSS channel that the item came from.
+
+
+class RSSIO(IOConfig):
+
+    _input: Optional[RSSInput] = None
+    _output: None
+
+    _sites: List[str]  # A list of sites to poll for RSS update events
+    _polling_every: int  # Total re-poll time for the sites in the _sites list
```
exachixkitsune: Presuming seconds?

src/mewbot/io/rss.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def __init__(
+        self, sites: List[str], polling_every: int, startup_queue_depth: int = 5
+    ) -> None:
+        super(Input, self).__init__()  # pylint: disable=bad-super-call
+
+        self._sites = sites
+        self._sites_iter = cycle(iter(self._sites))
+
+        self._polling_every = polling_every  # The time taken to poll ever site in sites once
```
exachixkitsune, line 256: Do we want to warn if this polling interval is very small? Not to enforce a min, but to let the user know if they've entered a bad value.

exachixkitsune:
Erroneous commit removed via reset
Errant branch is https://github.com/mewler/mewbot/tree/rss_experiment_errantcommit if anything major was lost.

src/mewbot/io/rss.py
exachixkitsune: ANYWAY
line 255 here has an errant ). Any change would want to remove it

exachixkitsune (commented):
Minor comments only; the other review covered everything major.

src/mewbot/io/rss.py
javajawa: I sneakily fixed the suggestion at some point :3

exachixkitsune (dismissed):
I am content from my point of view; the other comments remain.

src/mewbot/io/rss.py
ajCameron: > `asyncio.get_running_loop()` could even be determined at `__init__` time, to be honest

Loop isn't up at __init__ - but it can wrap it as a property - which may improve efficiency somewhat.

src/mewbot/io/rss.py
ajCameron: So - my thinking for this.

I considered a generic method, however some poking around revealed that there are lot of different ways which people have implemented the RSS standard. The current methods are more placeholders, with the aim to replace them with more sophisticated parsing methods when I have time to prod some of the weirder edge cases with a stick.

This might, or might not, be worth actually doing. Will give some more thought to it.

src/mewbot/io/rss.py
ajCameron: Indeed - added.

src/mewbot/io/rss.py
ajCameron: This makes sense - and seems part of making the project more user friendly in general. Will add.

src/mewbot/io/rss.py
ajCameron: Updated to seperate

src/mewbot/io/rss.py
ajCameron: It can be, and should.

Have added it to style guide notes (as I think this is generally preferable?)

src/mewbot/io/rss.py
ajCameron: ohh - yes. I had totally misunderstood. Corrected.

src/mewbot/io/rss.py
ajCameron: Updated to improve readability

src/mewbot/io/rss.py
ajCameron: That does, indeed, make more sense. Corrected.

src/mewbot/io/rss.py
ajCameron: Having looked at it both ways, I think it definitely does - and allows for greater flesxibility.

Updated.

src/mewbot/io/rss.py
ajCameron: Now updated to do so.

requirements.txt
ajCameron: Very sensible - adopted

src/mewbot/io/rss.py
ajCameron: Decided it was not worth the additonal LOC - if anyone needs to do custom extractors, they can add them later.

Kept the guid and pubDate methods, as those 
 - for guid, we need _something_ - so included a fallback method
 -  If the pubDate field is not filled then the source is producing very malformed RSS - which seemed worthy of a call to the log

exachixkitsune (dismissed):
Seems fine to me.

javajawa (dismissed):
Looks good asides from test coverage (and maybe cleaning up the giant commit message of doom)

exachixkitsune (dismissed):

ajCameron:
Issues with test coverage are resolved. Hopefully just one more round of approvals and can get it on mainline.

javajawa (approved):

PR #43, opened 2022-04-30
11 add testing harness

### LINKED ISSUE #11: Add testing harness

### LINKED ISSUE #27: Pylint version not functional on windows
> > it's erroring on pylint with a RecursionError; around colorama and asteroid

> Are those being pulled in by mewbot, or coming in from the global environment?

Good question.
So it appears that they are used by pylint / multiprocessing - pylint is spawning multiple processes, and so it's using multiprocessing, and that's failing because these are dependants...
I updated pylint and it failed.
Seems like, from looking at other github things, it's an issue with pylint and colorama specifically.
(https://github.com/PyCQA/pylint/issues/6012;
I am using:
pylint 2.13.5
dill 0.3.4 (in the traceback)
colorama 0.4.4)

__Originally posted by @exachixkitsune in https://github.com/mewler/mewbot/issues/22#issuecomment-1094336549__

exachixkitsune:
Here's hoping this all works properly...

This should be an automated test framework, with a workflow entity to make this go. I haven't actually tested it yet, but fingers crossed the workflow goes.
I am also expecting the test cases to fail after the merge happens due to code changes.

May well need the integration to occur before working out if it reports properly. I've never done this before. Time to see how it goes!

Closes #11 
Potentially fixes #27


exachixkitsune:
Also I've just realised the reporting for the workflow actions may want particular formats; can revise if necessary.

tests/common.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class BaseTestClassWithConfig(ABC):
+    # pylint: disable=R0903
+    # Too few public methods, because this is always used elsewhere
+    fname: str
+    config: ConfigBlock
+    doctype: ComponentKind
+    config_toload = True
+
```
javajawa: I love the concept of this class! Very nice way to load a component from YAML.

This comment covers two additional thoughts:
 - I find overriding properties to be an unclear way to give info to a super class
 - We could do some fun things with the type system

For the first, I'd recommend setting up some stuff in `__init__`:
```suggestion
class BaseTestClassWithConfig(ABC):
    """Caches configs and components for testing"""
    fname: str
    doctype: ComponentKind
    _config: Optional[ConfigBlock] = None
    config_toload = True

    def __init__(self, filename: str, kind: ComponentKind) -> None:
        self.fname = filename
        self.doctype = kind

    @property
    def config(self) -> ConfigBlock:
        self.cache_config()
        return self._config

    @property
    def component(self) -> Component:
        self.cache_config()
        return load_component(self.config)
```

For `ExampleHttpPostIOConfig`, this would make the class look more like:
```python
class ExampleHttpPostIOConfig(BaseTestClassWithConfig):
    def __init__(self) -> None:
        super().__init__("examples/trivial_http_post.yaml", ComponentKind.IO_CONFIG)

    @property
    def component(self) -> HTTPServlet:
        component = super().component
        assert isinstance(component, HTTPServlet)
        return component
```

The typing idea in the second bullet is around using the Generics in python typing, which might look something like:

```suggestion
T = TypeVar("T", bound=Component, covariant=True)


class BaseTestClassWithConfig(ABC, Generic[T]):
    _config: ConfigBlock
    _component: Optional[T] = None

    def __init__(self, filename: str, implementation: Type[T]) -> None:
        with open(filename, "r", encoding="utf-8") as config:
            self._config = next([document for document in yaml.load_all(config, Loader=yaml.CSafeLoader) if document["implement"] == implementation.__name__])

    @property
    def config(self) -> ConfigBlock:
        return self._config

    @property
    def component(self) -> T:
        if not self._component:
            component = load_component(self._config)
            assert isinstance(component, T), "Config loaded incorrect type"
            self._component = component
        return self._component
```

Which should mean that doing something like `BaseTestClassWithConfig("examples/my_example.yaml", mewbot.discord.DiscordIO)` will type check as returns in a `DiscordIO` without any extra work

javajawa: ^ Note: I wrote all that from memory, bug me if things don't actually work

tests/common.py
```diff
@@ -0,0 +1,43 @@
...
+    def cache_config(self) -> None:
+        if self.config_toload:
+            with open(self.fname, "r", encoding="utf-8") as config:
+                for document in yaml.load_all(config, Loader=yaml.CSafeLoader):
+                    if document["kind"] == self.doctype:
+                        self.config = document
+                        self.config_toload = False
```
javajawa: I'd invert the condition here to reduce indentation, and also exit out once we match an item
 ```suggestion
        if not self.config_toload:
            return
        with open(self.fname, "r", encoding="utf-8") as config:
            for document in yaml.load_all(config, Loader=yaml.CSafeLoader):
                if document["kind"] == self.doctype:
                    self.config = document
                    self.config_toload = False
                    return
```

File: tests/test_core.py [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+    ActionInterface,
+    IOConfigInterface,
+)
+
+# pragma pylint: disable=R0903
+#  Disable "too few public methods" for test cases - most test files will be classes used for
+#  grouping and then individual tests alongside these
+# pragma pylint: disable=R0201
+#  Disable "no self use" for functions. These functions will not be used internally as they are
+#  automatically called by pytest as it seeks and searches for tests.
```
javajawa: You can avoid having to do this be removing the class, and just having a list of functions in the file. As far as I can tell, having the tests in a class doesn't do anything in terms of the information pytest returns.

tools/lint.py
```diff
@@ -60,23 +36,11 @@ def run_tool(self, name: str, *args: str) -> subprocess.CompletedProcess[bytes]:
...
         arg_list.extend(self.folders)
 
-        run = subprocess.run(
-            arg_list, stdin=subprocess.DEVNULL, capture_output=self.is_ci, check=False
-        )
+        run_result = run_utility(name, arg_list, self.is_ci)
 
-        if run.returncode:
-            self.success = False
+        self.success = run_result["success"]
```
javajawa: This is technically a change in behaviour -- previously, we could never go from false to true, but this permits this.
When fiddling locally, I used
```python
self.success = self.success and (run_result.returncode == 0)
```
(I also took out the typed dict with the `success` property and just returned the completed process directly).

That said, looking at this function, there is the possibilty of extracting an abstract class from thsese two, so that success logic isn't duplicated...

tools/test.py
```diff
@@ -0,0 +1,113 @@
...
+                    this_line = int(split_output[1])
+                    this_err = split_output[2].strip()
+                    if this_err == "":
+                        this_err = "OtherError"
+                    yield Annotation(
+                        "error", this_filepath, this_line, 1, this_err, current_test
+                    )
+
+            if ("FAIL Required test coverage" in output) and ("not reached." in output):
+                yield Annotation("error", "tools/test.py", 1, 1, "CoverageError", output)
```
javajawa: The annotation probably wants to go against the file what was not covered?

exachixkitsune: The threshold for code coverage appears to be overall coverage rather than specific file coverage; thus there is no specific file to relate to.

tests/test_core.py
exachixkitsune: (Regarding fixing, it probably ends up as a convention thing; I personally would group tests into classes, particularly because that allows commonality with some configuration settings for some cases, and then the other cases match the same convention - but that's a Kitsune stylistic thing.)

The class containing the tests is written into output messages when the tests fail.

For example, if I simulate an error in the function `test_componentkind_interface_map_ioconfig`, it says: `TestComponent.test_componentkind_interface_map_ioconfig` and `FAILED tests/test_core.py::TestComponent::test_componentkind_interface_map_ioconfig - AssertionError:...`

rather than removing `test_componentkind_interface_map_ioconfig` from it's class, instead I get: `test_componentkind_interface_map_ioconfig` and `FAILED tests/test_core.py::test_componentkind_interface_map_ioconfig - AssertionError:...`

Which only really matters for visually seeing where errors are, rather than the automated output system in `tools/test.py`

ajCameron: I find grouping tests into classes helpful in a number of ways
 - It allows for (easier) automated test generation. This can be particularly useful when trying to optimize (later - "premature optimization" and all) when, for example, you have a newly written optimized component which you want to test with a bunch of other configs to make sure everything all works
 - Given this program's going to be all config driven, I suspect being able to subclass tests for slightly different configs will be really helpful
 - It allows (again, possibly just me personally) to derive useful stuff from them more easily (e.g. "this started life as a test suite. Now it's part of my optimization and profiling toolchain")

So yes! I am opinionated in favour of test classes.

tests/common.py
ajCameron: I second the above on the form - very nice!

tools/common.py
```diff
@@ -0,0 +1,68 @@
...
+    if is_ci:
+        print(f"::group::{name}")
+        sys.stdout.write(run.stdout.decode("utf-8"))
+        sys.stdout.write(run.stderr.decode("utf-8"))
+        print("::endgroup::")
+    else:
+        run.stdout = b""
+        run.stderr = b""
+
+    return {"success": success, "completedProcess": run}
```
ajCameron: Looks good to me!

ajCameron (commented):
Looks good to me!

ajCameron (dismissed):
If my comments seem terse, that's mostly because @javajawa said most of what I was thinking. All looks good from here.

tools/test.py
ajCameron: Perhaps split the difference and annotate files that are _particularly_ bad? E.g. files with no test coverage at all. But I think that's a feature which could be left till later.

tests/common.py
javajawa: So it turns out that `pytest` will ignore classes with `__init__`, but I have been able to do "load this specific object from a config file" which looks like

```python
class TestLoaderHttpsPost(BaseTestClassWithConfig[HTTPServlet]):
    config_file: str = "examples/trivial_http_post.yaml"
    implementation: Type[HTTPServlet] = HTTPServlet

    # Test this working
    def test_working(self) -> None:
        assert self.config["implementation"] == "mewbot.io.http.HTTPServlet"

    # Test that the loading is accurate
    def test_loading_component_type(self) -> None:
        assert isinstance(self.component, HTTPServlet)
```

javajawa:
@exachixkitsune I took the liberty of implementing my suggestion for the typed loader, and did some housekeeping.

exachixkitsune:
> @exachixkitsune I took the liberty of implementing my suggestion for the typed loader, and did some housekeeping.

So I see; looks good to me at the moment.

javajawa:
@exachixkitsune @ajCameron I think this is now "done" -- not sure I'm happy with what I've done with the history, but it does preserve the conversations in this thread pretty well.

ajCameron:
It looks good to me! I'll do a final scan over the code and then approve.

ajCameron (approved):
All looks good to me! Thanks for getting this done.

PR #48, opened 2022-06-18
Configure dependabot

javajawa:
Dependabot will automatically create merge requests to update packages with security alerts.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


exachixkitsune:
Only question: does `rebase-strategy` automatically make pull requests, or does it automatically do something else?

javajawa:
> Only question: does `rebase-strategy` automatically make pull requests, or does it automatically do something else?

Dependabot makes PRs by default, that flags tells it to automatically rebase them is something changes.

exachixkitsune (approved):
All seems good then!

PR #49, opened 2022-06-24
Added windows setup instructions

ajCameron:
Detailed instructions as to how to set up mewbot on a clean windows install.

- Suggested settings if installing python
- setting up a venv
- adding shortcuts to make venv activation easier
- Test running the linters and the tests
- adding mewbot to the PYTHONPATH so the user can run the examples
- running the examples

Setup, after following instructions, is tested and working.


exachixkitsune:
There is an alternative to adding to `PYTHONPATH`, which is to add:
`.\mewbot_venv\Lib\site-packages\mewbot.pth`
with a single line in it, being the path to your src directory.
(Say, you have mewbot in `C:\mewbot_dev`, you would have `mewbot.pth` being `C:\mewbot_dev\src`)

May be too complicate to include, but may be a neat alternative.

javajawa:
> There is an alternative to adding to `PYTHONPATH`, which is to add:
> 
> `.\mewbot_venv\Lib\site-packages\mewbot.pth`
> 
> with a single line in it, being the path to your src directory.
> 
> (Say, you have mewbot in `C:\mewbot_dev`, you would have `mewbot.pth` being `C:\mewbot_dev\src`)
> 
> 
> 
> May be too complicate to include, but may be a neat alternative.

I didn't know that, and I wonder if it is possible for the other platforms/OSes -- might make some things easier overall

ajCameron:
I definitely think it's worth including as an alternative - much more elegant that hacking on the activate script directly.

javajawa:
If it accepts relative paths (which it might, depending on the underlying implementation), we can even commit it to the codebase...

ajCameron:
Have included use of .pth to modify the interpreter path as the preferred method.

dev-docs/windows-dev-setup.md
```diff
@@ -0,0 +1,462 @@
...
+
+A path file is just a text file with extension `.pth` which include folders you want added to the python path - one per line.
+
+It must be placed on the path of the python interpreter your currently using. In this case, assuming you have the same paths as me, probably the best place to put it would be
+
+```shell
+C:\mewbot_dev\mewbot_venv\Lib\site-packages
+```
+
+Simple add a file named `mewbot.pth` here with the single line `C:\mewbot_dev\mewbot`. It's probably best to add it as a new file, instead of modifying one of the existing ones - as these are under system control and may be overwritten without warning.
```
exachixkitsune:
```suggestion
Simple add a file named `mewbot.pth` here with the single line `C:\mewbot_dev\src`. It's probably best to add it as a new file, instead of modifying one of the existing ones - as these are under system control and may be overwritten without warning.
```
Pretty sure this should be as suggestion.

dev-docs/windows-dev-setup.md
```diff
@@ -0,0 +1,462 @@
...
+...
+:END
+```
+
+so it won't be printed every time you activate the venv.
+Save and close all open terminals.
+Then open a new terminal, activate the venv, and check that `tools` can now be found by doing
+
+```shell
+(mewbot_venv) C:\Users\Adriann Ves Bloche>python
```
exachixkitsune: check; accidental name inclusion?

exachixkitsune (changes requested):
Minor comments in file;
I will say, it's hard to read the induvidual sections - I often scrolled passed and missed them and the bold headings were more obvious.
Maybe make easy bit a heading, or an underline or something?

dev-docs/windows-dev-setup.md
ajCameron: Should be connected - thanks for the catch

dev-docs/windows-dev-setup.md
ajCameron: Corrected

ajCameron:
Hi @exachixkitsune 

Made some changes to the linting/testing scripts so they should work on windows, as well as instructions as to how to run the .sh files on windows.

Hope this helps and resolves https://github.com/mewler/mewbot/issues/143

ajCameron:
Superceded by https://github.com/mewbotorg/mewbot/pull/189

PR #58, opened 2022-06-29
Added functionality to Discord input

ajCameron:
Added
- Input events when messages are edited and deleted (not just created)
- Input events when a new member joins a guild
- The ability to request historic messages on startup
- Example bots for all these (apart from the new member join functionality - need a good way to test this)
- Refactored the Discord Input event class names for consistency
- Refactored examples to isolate all the Discord ones in a single subfolder


examples/discord_bots/delete_warn_discord_bot.py
```diff
@@ -0,0 +1,106 @@
...
+    DiscordInputEvent,
+    DiscordMessageCreationEvent,
+    DiscordMessageDeleteInputEvent,
+    DiscordOutputEvent,
+)
+
+
+class DiscordAllCommandTrigger(Trigger):
+    """
+    Nothing fancy - just fires whenever there is a DiscordTextInputEvent - of any type.
```
exachixkitsune: Probably wants modifying if that's no longer the case? Or does the event trigger, and then it gets used on some instances?

exachixkitsune (dismissed):
Looks good, only one comment on the comments in the files.
Are there other tests we could add to tests_io_discord.py? It feels like it may be hard/impossible to do a full test in a discord, but can we emulate that behaviour? - not important for this maybe

examples/discord_bots/delete_warn_discord_bot.py
ajCameron: No - I think that's just a stale comment. Will update.

ajCameron:
I've been thinking about how to test Discord properly as well.

I need to finish hacking it together, but I think there's scope for a full test - no mocking required. My approach would be to create a throwaway (hidden) channel, rig the Input class so it ignores all events but the ones in that channel, and then run the tests - emptying the channel after each. (The user would need to feed the tests suite a valid discord token ... but that's probably not too much to as).

Will think on it some more, and might give it a crack.

examples/discord_bots/delete_warn_discord_bot.py
ajCameron: Updated examples and comments to reflect new variable names - some more refactoring of the examples for clarity.

ajCameron:
Refactored some of the examples for added clarity (i.e. making the doc strings actually reflect reality).

src/mewbot/io/discord.py
```diff
@@ -91,6 +160,54 @@ async def on_ready(self) -> None:
...
+            messages = await channel.history(limit=5).flatten()
+            past_messages.extend(messages)
+
+        # Sort the messages and put the last five on the wire
+        past_messages = sorted(
+            past_messages, key=lambda x: float(x.created_at.timestamp()), reverse=True
+        )
+
+        for _ in range(self._startup_queue_depth):
+            message = past_messages[_]
```
javajawa: I do not like using `_` as an actual variable.
Would something like a slice make this clearer?
```suggestion
        for message in past_messages[0:self._startup_queue_depth]:
```

ajCameron: Clearer and possibly faster!

javajawa (approved):

PR #72, opened 2022-08-03
Folder monitor

ajCameron:
- allows monitoring of a file - or folder - for changes which generate input events
 - allows files to be written out to a destination folder as the result of an output event
 - provides a prototype for the sort of comprehensive tests it would be good to have for all the IO methods
 - and an example of the expanded dev notes, for comment.
- might not even be broken anymore!


examples/file_system_bots/dir_input_monitor_bot.py
```diff
@@ -0,0 +1,110 @@
...
+            InputFileDirDeletionInputEvent,
+        }
+
+    @staticmethod
+    def produces_outputs() -> Set[Type[OutputEvent]]:
+        return set()
+
+    async def act(self, event: InputEvent, state: Dict[str, Any]) -> None:
+        """
+        Construct a DiscordOutputEvent with the result of performing the calculation.
```
exachixkitsune, line 104:
```suggestion
        Print a logging message
```

examples/file_system_bots/file_input_monitor_bot.py
```diff
@@ -0,0 +1,81 @@
...
+            InputFileFileDeletionInputEvent,
+        }
+
+    @staticmethod
+    def produces_outputs() -> Set[Type[OutputEvent]]:
+        return set()
+
+    async def act(self, event: InputEvent, state: Dict[str, Any]) -> None:
+        """
+        Construct a DiscordOutputEvent with the result of performing the calculation.
```
exachixkitsune, line 75:
```suggestion
        Print a logging message of the event is appropriate.
```

exachixkitsune (commented):
I cannot say I've reviewed the files here in any great detail; but if you're happy and have tested it works don't let me hold you up

PR #79, opened 2022-08-26
Expand PyTest and Sonar integrations

javajawa:
## Install pytest modules for more flexibility
    
- `pytest-asyncio` has been added for testing async code
- `pytest-xdist` added for parallelising tests
- `pytest-github-actions-annotate-failures` for automatically generating github annotations
    
The introduction of these modules changes how we run pytest. Coverage and results are now output primarily as XML alongside the terminal versions. Tracking coverage is not enabled by default to allow for fast testing with parallelisation
    
Tests are run with 'new' first (i.e. not in the result cache), and a test taking more than a second is reported as a potential issue

## Switch Sonar reporting to an active mode
    
Rather than have Sonar automatically scan the repo based on a webhook, we instead explicitly upload the state of a request into it. This method allows us to also upload information like the coverage report, and the reports from the linters that Sonar is configured to use.
    
The ToolChain code in the tools module has been updated to output all toolchain output to files in `reports/`.


javajawa:
I really like that Sonar does the coverage calculation based on the lines you have changed -- so 100% of the code I have edited has code coverage, so it shows 100% coverage on the PR info

CONTRIBUTING.md
```diff
@@ -76,10 +76,14 @@ such as links to decisions about the structure of the code and notes features th
 The use of comments to note problems or future fixes is discouraged, and merges containing them will not be approved.
 Instead, create an issue to track the required change.
 
-<!--
 ## Testing
-This section is to be drafted as part of #11 "Add testing harness"
--->
+
+Testing an application framework which primarily deals with network interactions is always a difficult problem.
+The design of mewbot using a number of components that communicate indirectly through dataclasses partially
```
ajCameron: Grammar reads a little weird here - "uses"?. Perhaps expand slightly to add further clarity. Suggest instead something like

"The design of mewbot uses a number of components which communicate by passing dataclasses to each other. This assists with testing by allowing us to craft inputs which can then be injected to produce verifiable outputs."

tools/lint.py
```diff
@@ -5,9 +5,17 @@
 from typing import Generator, Set
 
 import os
+import sys
 import subprocess
 
-from tools.common import Annotation, ToolChain
+try:  # pylint: disable=duplicate-code
+    from tools.common import Annotation, ToolChain
+except ImportError:
```
ajCameron: Nice addition - might be an idea to have reminder at the end of the test run as well - if we're hitting this the user probably is trying to set up a dev environment and something has gone wrong - which they should probably fix.

tools/test.py
```diff
@@ -18,73 +19,68 @@
...
+            "--new-first",  # Do uncached tests first -- likely to be more relevant.
+            "--durations=0",  # Report all tests that take more than 1s to complete
+            "--durations-min=1",
+            "--junitxml=reports/junit.xml",
+        ]
+
+        if not self.coverage:
+            args.append("--dist=load")  # Distribute between processes based on load
+            args.append("--numprocesses=auto")  # Run processes equal to CPU count
+            args.append("--maxprocesses=16")  # Cap at 16 processes
```
ajCameron: Seems a reasonable mumber, but why the cap?

ajCameron:
Set of very reasonable upgrades! Small notes to the docs - otherwise very happy to pass it as is.

tools/lint.py
javajawa: I'll be honest I did this mostly so `./tools/lint.py` works, which I should probably do in my my config -- I remember in the windows work there was talk about essentially an environment file which could set this up properly so `tools/` is known to be the root of the `tools.` package...

tools/test.py
javajawa: I have no good reason!

tools/lint.py
exachixkitsune: The thing concerning the `.pth` file is discussed in https://github.com/mewler/mewbot/pull/49

.github/workflows/test.yml
```diff
@@ -9,7 +9,42 @@ on:
     types:
```
exachixkitsune: Do we want to still call this test.yml or would another name like workflows.yml work?

README.md
```diff
@@ -70,8 +70,16 @@ python3 -m examples examples/trivial_socket.yaml
...
 #### Running the tests and linters
 
-You can run the linters via the convenience script `tools/lint.py`.
-
-You can run the test framework using the convenience script `tools/test.py` (via `python -m tools.test`). This outputs html details of code coverage in `.coverage_html`.
-
-> :warning: However, as of this commit, no tests exist.
\ No newline at end of file
+You can run the linters via the convenience script `tools/lint.py` (or, more
+correctly, `python3 -m tools.lint`).
```
exachixkitsune: If we're saying "more correctly, thing" should the more correct option be as the suggestion?
```suggestion
You can run the linters via the convenience script `python3 -m tools.lint`
(or `tools/lint.py`).
```

README.md
```diff
@@ -70,8 +70,16 @@ python3 -m examples examples/trivial_socket.yaml
...
+This runs the auto-formatter `black`, two opinionated linting tools in `flake8`
+and `pylint`, and the `mypy` type checker in strict mode.
+
+You can run the test framework using the convenience script `tools/test.py` (or, more
+correctly, `python3 -m tools.test`).
```
exachixkitsune:
```suggestion
You can run the test framework using the convenience script `python3 -m tools.test`
(or `tools/test.py`).
```

CONTRIBUTING.md
```diff
@@ -76,10 +76,14 @@ such as links to decisions about the structure of the code and notes features th
...
+helps with this, allowing us to craft inputs and verify outputs.
+
+Tests are performed using the `pytest` testing framework, with tests located in the `tests` folder.
+Merge requests should not reduce the overall test coverage of the framework, as calculated by Sonar
```
exachixkitsune, line 86: SHOULD or MUST? Presumably exceptions must be agreed in the PR?

sonar-project.properties
```diff
@@ -0,0 +1,20 @@
...
+sonar.sourceEncoding=UTF-8
+
+sonar.coverage.exclusions = examples/**/*
+sonar.cpd.exclusions = examples/**/*
+
+sonar.links.scm=https://github.com/mewler/mewbot
+
+sonar.python.version=3.9, 3.10, 3.11
+sonar.python.coverage.reportPaths=reports/coverage.xml
+sonar.python.xunit.reportPath=reports/junit.xml
```
exachixkitsune, line 18: are we still making junit.xml files?

src/mewbot/api/registry.py
```diff
@@ -44,6 +44,7 @@ def __call__(  # type: ignore
             raise TypeError("Attempting to create a non registrable class")
 
         obj: Any = cls.__new__(cls)  # pylint: disable=no-value-for-parameter
+        # pylint: disable=attribute-defined-outside-init
```
exachixkitsune: out of curiosity, is there a reason not to fix this?
or is it likely to be in `abc.ABCMeta`?

exachixkitsune (dismissed):
only minor comments

tools/lint.py
javajawa: I am coming to the conclusion that this was a really neat hack, but I should just fix the documentation to always use `python -m tools.*` and not work around (my own) environmental issues

README.md
javajawa: Merging this thought with https://github.com/mewler/mewbot/pull/79#discussion_r956585798
I wrote that import trick to make it so the direct exec method worked, as that is the least surprising, but honestly having two subtly different ways to run the code isn't great.

sonar-project.properties
javajawa: We are! (we are not processing them for annotations in the workflow, because the new plugin does that during the test run, but we are uploading them to sonar)

src/mewbot/api/registry.py
javajawa: This is due to the typing in this magical method being very weird; logically the line abvoe we know that `obj` is of type `Component`, but asserting that in any way causes different issues. This one I have suppressed for two reasons: firstly, it is the one which the error is most 'true' for (init is not called for another 10 lines or so), secondly it has only just appeared as a detectable thing.

Frankly, this level of messing with magic methods kind of falls outside of the linter's domain

CONTRIBUTING.md
javajawa: I actually don't know how Sonar configures the gate, so I didn't want to write something stricter than what the tests say

tools/test.py
javajawa: ...I was going to remove it in my rebase. will do so if there's any other changes.

.github/workflows/test.yml
javajawa: Maaaayyybbeee.
I'm not sure what usefully encapsulates the purpose of the file.

`review`, with `Automated Review Pipeline` as the title?

src/mewbot/api/registry.py
javajawa: @ajCameron found a possibly neater fix for the linting but which masks the issue a little more: https://github.com/mewler/mewbot/pull/83/files#diff-2581cf6becc21d8526e55750bf2c59d4c79f271279474ae15b9dbb51158aaca8R21

javajawa: No longer a part of this PR

.github/workflows/test.yml
javajawa: Done

tools/test.py
javajawa: Removed

ajCameron (approved):

PR #83, opened 2022-08-31
Needed to patch in the methods to the discord client

ajCameron:
- Methods where not being called after the client had been split out
 - This seemed to be the best way which was acceptable to the linters

### DIFF
--- a/src/mewbot/api/registry.py
+++ b/src/mewbot/api/registry.py
@@ -18,6 +18,7 @@ class ComponentRegistry(abc.ABCMeta):
     """MetaType which reg"""
 
     registered: List[Type[Any]] = []
+    uuid: str  # This property is unused but fixes some linting issues in __call__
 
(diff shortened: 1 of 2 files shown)


ajCameron:
Seems to have happened after the changes to accout for the pycord-2.0 release

src/mewbot/api/registry.py
```diff
@@ -18,6 +18,7 @@ class ComponentRegistry(abc.ABCMeta):
     """MetaType which reg"""
 
     registered: List[Type[Any]] = []
+    uuid: str
```
javajawa: I... am a little concerned that this works

javajawa (dismissed):

src/mewbot/api/registry.py
ajCameron: Honestly? So was I!

I wrote it, thought it would fail, then checked to see if the linters complained, and they did not.

Namespaces are a hell of a drug.

PR #84, opened 2022-09-02
Add description to the registry class

javajawa:


src/mewbot/api/registry.py
```diff
@@ -15,7 +15,28 @@
 
 # noinspection PyMethodParameters
 class ComponentRegistry(abc.ABCMeta):
-    """MetaType which reg"""
+    """ComponentRegistry is a AbstractBaseClasses MetaClass which instruments
+    and modifies class definition/creation to automatically record and classify
```
ajCameron: Does not seem to modify them - but does instrument them.

src/mewbot/api/registry.py
```diff
@@ -15,7 +15,28 @@
...
+    classes that implement a given set of base classes.
+
+    This registry is designed specifically for use with components using the
+    YAML representation and bot loader, and performs certain property
+    manipulations in line with that design.
```
ajCameron: Suggest adding - (E.g. setting the UUID of the classes).

ajCameron (dismissed):

ajCameron (approved):

PR #85, opened 2022-09-02
Add IO for writing to local file system

javajawa:
The IOConfig only supplies an Output, which can create and delete files, including writing or appending content.

### DIFF
--- a/plugins/mewbot-io-file-storage/requirements-dev.txt
+++ b/plugins/mewbot-io-file-storage/requirements-dev.txt
@@ -2,7 +2,4 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
-# Used for reading configuration in mewbot.loader
-PyYAML~=6.0
-
-aiohttp>=3.8.0,<3.9.0  # Waiting for py-cord to upgrade
+types-aiofiles~=23.0
--- a/plugins/mewbot-io-file-storage/requirements.txt
+++ b/plugins/mewbot-io-file-storage/requirements.txt
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
-# Used for reading configuration in mewbot.loader
-PyYAML~=6.0
-
-aiohttp>=3.8.0,<3.9.0  # Waiting for py-cord to upgrade
+# File Watcher Modules
+aiofiles~=23.0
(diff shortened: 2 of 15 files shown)


javajawa:
I found some notes:
- Add test for creating dirs
- Add test for binary write
- Add test for attempting to write to `../` or `/abs` style paths
- Add test for non-existent output dir
- Add test for non-initialised output dir

src/mewbot/io/file_storage.py
```diff
@@ -0,0 +1,162 @@
...
+@dataclasses.dataclass
+class CreateDirectoryOutputEvent(FSOutputEvent):
+    """Creates a directory at the given path (and any parent directories)"""
+
+
+@dataclasses.dataclass
+class WriteToFileOutputEvent(FSOutputEvent):
+    """
+    Writes the given contents to the specified file.
+
```
ajCameron: Suggest adding something like "within the designated _output_path" - just to emphasize that arbitary file creation is not supported.

src/mewbot/io/file_storage.py
```diff
@@ -0,0 +1,162 @@
...
+    file_contents: Union[bytes, str]
+    append: bool = False
+    may_overwrite: bool = True
+
+
+@dataclasses.dataclass
+class DeleteFileOutputEvent(FSOutputEvent):
+    """
+    Deletes a file with the given name.
+    """
```
ajCameron: Again suggest adding "inside the designated _output_path"

ajCameron (dismissed):

ajCameron:
All looks good apart from some very minor doc suggestions.

exachixkitsune:
Does this pull remain valid, and/or want to wait until #120 is done to comply with that system?

javajawa:
> Does this pull remain valid, and/or want to wait until #120 is done to comply with that system?

Merging will wait for whatever change actually implements the plugin system (that's where my series of merges are slowly approaching base off of #127), but moving it to use the plugin system is pretty simple.

ajCameron:
What the...

ajCameron:
Okay. Name clash between a remote branch and a local branch.

Annoying. But fixable.

javajawa:
Handled in https://github.com/mewbotorg/mewbot-io-file-storage/pull/3

PR #88, opened 2022-09-09
Manager experiment

ajCameron:
Sketch of the eventual form of the manager is now complete.

Additional capability needed to be added, but a review of the direction would be appreciated.


dev-docs/component-dev-notes.md
```diff
@@ -0,0 +1,30 @@
+### Errors
+
+Components are intended to interface with the registry system - they are registratable objects.
+The intent of this is that the system knows what objects have been registered so it can present to the user on an API level all the components which can be included in a bot configuration.
+However, setting this up does mean adding some definitions in a couple of different places.
+
+### If you are changing a declared API
```
exachixkitsune: Probably want to consider a "if you're changing an API, increment the verno as per this scheme"; or at least a reference here to the other document which says how that's done. An api-dev-notes.md?

exachixkitsune: Do we want to say "<these> changes do not require an increment to the api". Probably stuff along the lines of "you are purely adding new features not modifying old ones"

exachixkitsune (dismissed):
Obviously needs the conflict fixing.
Code not reviewed in great detail, but not seen any glaring issues.

dev-docs/component-dev-notes.md
ajCameron: This seems clearer - updated

ajCameron: I have added a reference to api-dev-notes.md - will construct.

PR #91, opened 2022-09-14
please ignore this pull request - and sorry for the forethcoming spam

ajCameron:
The linters are breaking when run remotely - but not when run locally.

Additionally - weirdly - I have not been able to reproduce in a clone of the repo.

So working through manager changes and trying to determine what breaks the linters.


ajCameron:
Okay. Something in tools is breaking things.

PR #114, opened 2022-12-12
Groupped dependabot updates

javajawa:
Updates:
- setuptools from ~=65.4 to ~=65.6
- pytest-asyncio from ~=0.19.0 to ~=0.20.3
- py-cord from ~=2.1.3 to ~=2.3.2
- black from ~=22.8 to ~=22.12
- trufflesecurity/trufflehog from 3.13.0 to 3.19.0


ajCameron (approved):
I think I now get why that was causing issues for the type linting.

PR #120, opened 2022-12-27
[Draft] Prototype for the plugin system

ajCameron:
- Plugin system with the following features
  - Automated plugin discovery and loading
  - Allows declaration of src code code - which will be fully linted if declared
  - Allows declaration of examples for inclusion
  - Allows declaration of tests for inclusion into the test harness

Included as examples
 - reddit client
 - example day of the week conditions
 - a complete (basic) dice roller for discord


javajawa:
To summarise a conversation about this change elsewhere:

The functionality of these change can be seen as two parts:
 - Creating tooling to find an load mewbot components
 - Creating tooling to find and list mewbot source code, test, and example directories

In order to do so, plugin authors will need to add a block of boiler plate code. We already have similar functionality for identitfying components via the Registry system, but that requires knowlege of what modules to load. Looking at how pluggy works internally, it uses some parts of [setuptools entry points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) to determine what installed packages present relevant components. This requires a one line change in `setup.py` for the module in order to work. ([source](https://docs.pytest.org/en/4.6.x/_modules/pluggy/manager.html#PluginManager.load_setuptools_entrypoints)).

This gives us the list of modules, and thus the list of components, essentially 'for free' -- no need to declare extra metadata functions, just mark yourself as a mewbotv1 plugin in the setup config.

`setuptools` also has the [test_suite](https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-test-suite) which can be used for test detection; we will reclassify examples as being part of the test "suite", but not actually tests cases. This will mean additional work on the updated example finder, which will have to look at all the test source locations, and find exmaples within them.

exachixkitsune:
This sure seems to be a bunch of changes, some of seem to be conflicting with other things (like #129/#130/#134 at least). Is it still valid for review?

javajawa:
> This sure seems to be a bunch of changes, some of seem to be conflicting with other things (like #129/#130/#134 at least). Is it still valid for review?

Whoops, missed this comment -- broadly "no", this was superseded by #127, which was then broken down into smaller changes to make it easier to review.
I'm going to hole the merge here open to make sure we don't lose anything important as I do repeated rebases and rewrites (for example, #127 I think loses the dice roller)

javajawa:
Covered in other PRs (#129, #130, #131 ...)

PR #127, opened 2023-02-04
[Draft] Simpler Plugin Management

javajawa:
This will be broken down into a smaller number of merges (or at least commits) broadly covering:

 - Adding all the correct license information
 - Merging the `mewbot.core` and `mewbot.config` into a single module for packaging
 - Moving around the test helpers to form the `mewbot.test`  package
 - Adding the tooling for populating the Registry from the setuptools metadata
 - Adding the tooling to update the package diagram drawing script
 - Moving the in-tree plugins into the plugins folder, and adding the install_modules tool
 - Adding in the various new plugins
 - Publishing all the packages

Work that is still outstanding

 - Examples are currently not a working thing. Ideas have gone around for this, but my current favourite is just a data file called `examples.yaml` which contains the path or the actual example yamls. The UI will be updated to use ncurses, making it easier to display things like descriptions. They could also be implemented using the same YAML pattern as the actual components, and therefor have properties that could be set in the loader (e.g. API tokens).
 - The Reuse license linting needs to be properly set up in the codebase, tools, and CI pipeline so it detects files without license metadata.
 - A tool should be added to generate a useful `PYTHONPATH` variable that includes the appropriate things. This will fix the problem where I can not both pylint and mypy to pass due to the "duplicate" `tests` modules.

Work that will need to happen elsewhere

 - We should set up a template repo containing the bare bones of a mewbot IO plugin, with instructions on how to get started. It should only need to declare `mewbot-api` as a dependency

---

Here follows all of the commit messages that got squashed into the PR

---

Even DRYed

Manager interface stabilized

 - Test plugin system created
 - Example plugins written and can be installed with setup.py
 - Currently offers component support
 - Spec and loader framework added

Initial draft of the conditions and dev plugin complete

 - Manager exists in basic form but needs interface work
 - plugins can declare components to be exposed by the plugin manager
 - plugins can declare where their src is - for linting purposes
 - Example plugins being brought up to main code base standards
 - fallbacks added to tools in case pluggy not found

Updates required to make sure all new plugins correspond to coding conventions

 - Mostly for demonstration purposes

Test plugin integration still in development

 - some changes to the tests appear to be necessary
 - due to confusing being caused by the tests use of relative paths to load examples
 - overcoming this by adding a method to classes to allow fetch of a named example
 - this should also simplify integration of plugin examples

Preliminary work towards splitting off tests as their own module

 - currently trying to wrangle coverage to produce utile results

This solution currently works without --cov, but breaks badly with it

 - moving to a mewbot-tests module level approach.

Prototype of the plugin test system is now working

 - allows you to declare src code locations for --cov purposes
 - allows you to declare test locations
 - tests will be picked up and run - you can invoke them through the old mechanism, as before
 - tests from plugin examples included and running

Basic plugin infrastructure is now present

 - New plugins can be declared with
   - sources
   - examples
   - tests
 - Linting of all files containing python code is supported
 - duplicate code has been minimized as far as possible
 - all tests are passing
 - all type information present and correct

Added an example chooser option

 - To support the fact that the examples can be declared in multiple locations

Corrected error to setup.py

 - Classifiers where not being set correctly

Update __main__.py

Interface for plugin manager now - largely - exists

 - Can retrieve objects by name
 - Can also retrieve a list of available options.
 - Test backing is provided.

Additional tests to boost coverage

Removing stray file

First attempt to produce working install and lint flow

Reverting review.yaml

Single change - attempting to manually install tests plugin

Path sep confusion

A path is provided - but not seen as such

Directory confusion?

Mewbot itself needs to be installed

 - or pip will not recognize the dep as met

That was not the proper way to do it

Corrections for linting - final tests for review.yaml

Attempting to fix an requirement install problem in the example plugins

Typo when copying the requirements install lines

Interestingly some errors appeared when linters not run locally

Adding install steps for tests step

Port freeing issue in the github test runner?

 - resolved by manually making sure the IoHTTP sockets do not clash

Test sonar behaviour

Update the component concept

- Add a helper script to automatically detect and install plugins in the current tree
- Add the install requirements to various package's setup.py definitions
- Add a helper function to the registry to load in lpugins
- Add a demo test script to check the above works

A similar change is needed in the example finder still, and then the pluggy logic can be removed.

Improve module loading

Updates...

Noodling on things


javajawa:
Covered in other PRs (#129, #130, #131 ...)

PR #128, opened 2023-02-05
Update code with latest lints

javajawa:
- black 23 remove initial blank lines in new code scopes
- pylint 2.16 correctly handles meta-class argument names
- setuptools moved from requirements to requirements-dev


ajCameron (approved):
Will have to work out a hack to make pylint happy with the registry changes - but that is my OCD and not, actually, an operational piority.

PR #129, opened 2023-02-05
Merge the mewbot.core and mewbot.config modules

javajawa:
This change is in preparation for deploying multiple packages. These two modules would form the `mewbot-core` package; both are very small, and as such it seems easier to combine them into a single module.


javajawa:
I'm going to recommend reviewing this one as three separate commits, as the overall change is enough for git to lose track of files being moved around.

ajCameron (dismissed):
Tested working on windows (though not from a clean setup).

ajCameron (approved):
Will need to do some convenience tooling for windows

PR #130, opened 2023-02-05
Merge Python source locations to a single directory

javajawa:
In order to make environment management easier, the python code in `examples` and `tools` has been merged into `src`.

The examples loader has been made into an `examples` module. The `tools` module has been renamed to `mewbot.tools`, and added to the main tree. Both have been excluded from the setuptools package.

All commands in `tools` have been replaced with a simple POSIX shell script that handles setting `PYTHONPATH` automatically. The documentation has been updated to reflect this and also encourage use of this script as part of environment setup.


ajCameron (approved):
Looks good to me! Looking forward to seeing the final form.

PR #131, opened 2023-02-05
Add SPDX license information to all existing files.

javajawa:
This removes my own name from a bunch of license info, and is a step in the right direction for having (`reuse`)[https://reuse.software/] be a functional part of the toolchain.

Marked as draft because:
 - Docs should probably be a CC-based license, as they are not 'code'
 - The example YAML files might also benefit from that.

### DIFF
--- a/tools/reuse
+++ b/tools/reuse
@@ -5,14 +5,6 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 COPYRIGHT="Mewbot Developers <mewbot@quicksilver.london>"
-
 LICENSE="BSD-2-Clause"
 
-find . \
-	-path './.git' -prune -o \
-	-path './.reuse' -prune -o \
-	-path './LICENSES' -prune -o \
-	-path './venv' -prune -o \
-	-type f \
-	-exec \
-reuse addheader --merge-copyrights --copyright "$COPYRIGHT" --license "$LICENSE" --skip-unrecognised '{}' +
+reuse annotate --merge-copyrights --copyright "$COPYRIGHT" --license "$LICENSE" --skip-unrecognised --skip-existing --recursive .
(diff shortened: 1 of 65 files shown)


tox.ini
```diff
@@ -1,3 +1,7 @@
+; SPDX-FileCopyrightText: 2021 - 2023 Mewbot Developers <mewbot@quicksilver.london>
```
exachixkitsune, line 1: File has two SPDX segments

tools/reuse
```diff
@@ -5,14 +5,6 @@
...
 
-find . \
-	-path './.git' -prune -o \
-	-path './.reuse' -prune -o \
-	-path './LICENSES' -prune -o \
-	-path './venv' -prune -o \
-	-type f \
-	-exec \
-reuse addheader --merge-copyrights --copyright "$COPYRIGHT" --license "$LICENSE" --skip-unrecognised '{}' +
+reuse annotate --merge-copyrights --copyright "$COPYRIGHT" --license "$LICENSE" --skip-unrecognised --skip-existing --recursive .
```
exachixkitsune, line 10: Query: Does this apply evenly across every file or just code files?

exachixkitsune (commented):
Some minor.
Also does author= need changing in setup.py?
Also would a doc in the devdocs to indicate what license should be _nominally_ used for each file?

tools/reuse
javajawa: Every file, but the tool is doing file type stuff to work out how to add a header.

javajawa:
> Also does author= need changing in setup.py?

oops, that has ended up in the next set of changes because I rewrote `setup.py` in that one.

> Also would a doc in the devdocs to indicate what license should be _nominally_ used for each file?

That would be a sensible idea. Also possibly `contributing.md`.

exachixkitsune:
I am content with the changes but as it relates to licensing I feel @ajCameron should also have input.

PR #134, opened 2023-02-08
Merge Python source locations to a single directory

javajawa:
In order to make environment management easier, the python code in `examples` and `tools` has been merged into `src`.

The examples loader has been made into an `examples` module. The `tools` module has been renamed to `mewbot.tools`, and added to the main tree. Both have been excluded from the setuptools package and coverage requirements.

All commands in `tools` have been replaced with a simple POSIX shell script that handles setting `PYTHONPATH` automatically. The documentation has been updated to reflect this and also encourage use of this script as part of environment setup.


javajawa:
This is an updated clone of #130; for some reason when 130 got added into the 129 branch, the linting started rejecting it, so they have been re-split

.github/workflows/review.yaml
```diff
@@ -75,9 +73,7 @@ jobs:
         run:  pip install -r requirements-dev.txt
 
       - name: Run Tests
-        run: python3 -m tools.test --cov
-        env:
-          PYTHONPATH: "./src:."
+        run: ./tools/test
```
exachixkitsune: Should this still have --cov in order to generate coverage?

exachixkitsune (dismissed):
Content the testing seems to work still; though coverage is missing, so isn't reported in sonar. If that's deliberate, -> ok merge on.

exachixkitsune:
(will say none of the shell scripts work on windows but that doesn't seem to stop things working)

.github/workflows/review.yaml
javajawa: Oh. *that's* what broke #129/#130

javajawa:
```suggestion
        run: ./tools/test --cov
```

exachixkitsune (approved):

PR #136, opened 2023-02-12
Build a new setup system

javajawa:
## Finish refactoring code into packages

- Remove examples/ from coverage reports
- All code in examples/ has been moved to src/mewbot/examples
- Move test utilities into mewbot.tests package
- Split requirments file per package
- Export InputQueue and OutputQueue from mewbot-api
  
## Configure setuptools to work with multiple packages
    
This change adds support for building and installing the mewbot subpackages mewbot-core, mewbot-api and so forth.
It works by dynamically generating the setuptools configuration based on the arguments given to setup.py.
    
Each of the named packages will include the module `mewbot.[name]` (or just `mewbot` for the `mewbot` package), and will pull the dependencies from the `requirements-[name].txt` file. Packages that are not `mewbot-core` will also depend on `mewbot-core`.

Each of the mewbot packages will be automatically published to the PyPI repositories when a GitHub release occurs


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,52 @@
...
+          rm -Rf build; python setup.py 'core' sdist bdist_wheel
+          rm -Rf build; python setup.py 'api'  sdist bdist_wheel
+          rm -Rf build; python setup.py 'io'   sdist bdist_wheel
+
+      - name: "Archive installation artifacts"
+        uses: actions/upload-artifact@v2
+        with:
+          name: builds
+          path: dist/*
+
```
ajCameron, line 47: It's not perfectly clear to me how the version number for the package is being set before upload?

README.md
```diff
@@ -45,8 +54,9 @@ enforce the code style and linting requirements of this project.
 
 The recommended way to set this project up is using python 3.9 (or higher) with
 a standard `venv` setup.
-The project uses a "src-dir" layout, so you will need to set up `PYTHONPATH`
-with at least the `src/` directory for Python to locate the modules.
+The project uses a "src-dir" layout; for Python to be able to locate the modules, 
+you will need to set up the `PYTHONPATH` environment variable with at least the `src/` directory.
+
 For convenience in POSIX-like systems, an importable script `./tools/path`
 is provided that will return the appropriate path.
```
ajCameron, line 62: Suspect all the docs are going to need some attention - but that can wait until this PR is on mainline

requirements-api.txt
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: 2021 - 2023 Mewbot Developers <mewbot@quicksilver.london>
```
ajCameron, line 1: I'll get you access to this

setup.py
```diff
@@ -1,35 +1,102 @@
...
-)
+"""
+A hacked up setup.py which will install all of `src` as one package locally,
+but is used to build multiple packages for publishing.
+
+This script works only in certain edge cases, namely when setuptools is using
+the package name as the temporary directories (otherwise, multiple invocations
+will cause cross-talk with modules being included in packages they should not
+be in).
+
```
ajCameron, line 11: I'll think on some way to make this more elegant - but it's working for now

src/mewbot/io/rss.py
```diff
@@ -29,9 +29,7 @@
 import aiohttp
 import feedparser  # type: ignore
 
-from mewbot.core import InputEvent, OutputEvent
-
-from mewbot.api.v1 import IOConfig, Input, Output
+from mewbot.api.v1 import IOConfig, Input, Output, InputEvent, OutputEvent
```
ajCameron, line 33: Much better!

ajCameron (dismissed):
Looks good locally - tested install worked and had a mess around with it.

However the github actions seem to be all hanging remotely.

.github/workflows/deploy.yaml
javajawa: Currently...it's not! It's stuck at [0.0.1 forever](https://github.com/mewler/mewbot/pull/136/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R50) right now.

<details>
<summary>Very rambling thoughts on version management in Git</summary>

So, Git flows honestly have a bunch of issues with doing versioning well, especially when you have CI/CD going on.
The base logical way to deal with it is:
 - Each merge request adds an item to an "unreleased" section of the changelog
 - Have a merge which bumps a version number and sets that as the top version in the changelog
 - Tag the result of that merge
 - Release from the tag.

That gives you your publishable versions. But, if your building an arbitary commit, how do you version that? Well, `git describe` is a great way for generating version strings, except it is looking at commits.
To me, the most accurate logic is:
1. Use `git describe` to get the nearest tag
1. Use `git rev-parse [tag]:src` to get the tree hash of the source code at the tag
1. Use `git rev-parse HEAD:src` to get the local tree hash
1. If the two hashes match, functionally you are at the same package (just a different history or documentation changes
1. If the two don't match, your version is `[tag]+[HEAD tree hash]`

GitHub has a built in releases feature that does...some of this. However, it doesn't have any support for automatically updating changelogs in the repository (which I find disappointing), and the web UI for releases only considers things one release at a time.

In conclusion, version management I forsee primarily using git features, possibly with some cool logic, and github will be supplementary
</details>

setup.py
javajawa: Most of the problem is that setuptools doesn't honour it's own temporary directory directives for certain commands >.< I had a much nicer version that just added some flags, but it was always building into `build/mewbot` regardless of what options I used.

.github/workflows/deploy.yaml
ajCameron: Ahhh. Well. That sounds like a nightmare for another time.

ajCameron (approved):
Looks good!

.... Now just need to update all the branches and clear the decks on pull requests!

PR #142, opened 2023-02-24
Add MacOS, Windows, and python 3.11 to auto review

javajawa:
GitHub is now configured to lint on python versions 3.9 through 3.11 on the current versions of Windows Server, MacOS, and Ubuntu Linux.

### DIFF
--- a/.github/workflows/review.yaml
+++ b/.github/workflows/review.yaml
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 name: Auto-Review
+run-name: "Automated code review for \"${{ github.head_ref }}\""
 
 on:
   push:
@@ -18,6 +19,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  # Which version of the reports Sonar receives.
+  DEFAULT_OS: ubuntu-latest
   DEFAULT_PYTHON: 3.9
 
 jobs:
@@ -53,11 +56,12 @@ jobs:
 
   test:
     name: Test Python code
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
-        version: ["3.9", "3.10"]
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        version: ["3.9", "3.10", "3.11"]
       fail-fast: false
 
     steps:
@@ -77,14 +81,26 @@ jobs:
         run:  pip install -r requirements-dev.txt
 
       - name: Run Tests
+        shell: bash
         run: ./tools/test --cov
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v3
         if: always()
         with:
-          name: test-results-${{ matrix.version }}
+          name: test-results-${{ matrix.os }}-${{ matrix.version }}
           path: reports
+
+  tests-passed:
+    name: Tests Passed
+    runs-on: ubuntu-latest
+
+    needs:
+      - test
+
+    steps:
+      - name: Mark all test runs as successful
+        run: /bin/true
 
   sonarcloud:
     name: SonarCloud
@@ -112,7 +128,7 @@ jobs:
         uses: actions/download-artifact@v3
         if: always()
         with:
-          name: test-results-${{ env.DEFAULT_PYTHON }}
+          name: test-results-${{ env.DEFAULT_OS }}-${{ env.DEFAULT_PYTHON }}
           path: reports
 
       - name: SonarCloud Scan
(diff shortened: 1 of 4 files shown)


javajawa:
The tests are no currently running on the Windows runners as they are not picking up `PYTHONPATH` properly from the `./tools/path` script (presumably due to the difference between how Linux and Windows environment variables are working. 

Workflow specific fixes for this would include adding it to the Workflow's environment variables, of building a verison of the `./tools/path` script that perform an [update of the GitHub environment](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files). _However_, my preference would be for a more general fix for Windows users of the platform.

exachixkitsune:
Currently, doesn't seem like, without shenanigans, the files will run in windows. In particular, getting sh to run requires git to have been set up in a particular way.

I am considering one of the following options:

1. Making duplicate files for tools which are powershell files, forking the github actionsso it runs different commands for windows
2. Adjusting the github actions to run python directly not via scripts; embedding path modification in there.
3. Changing the tools scripts to be bash or something else transcompatible and making everything work with that.
4. Mandating WSL is used.

Then the path script can be made to do environment variables correctly, correctly set src and test on the path, and then continue on.

Thoughts accepted. My preference is number 3 but it may break your toolchains?

javajawa:
My preference is against setting the paths as a fixed string anywhere because that will itself be a python script in an planned update (see [[1]](https://github.com/mewler/mewbot/blob/f77725bb168df62cd9c8db1f8f35a9d2a6ec88ae/tools/path) and [[2]](https://github.com/mewler/mewbot/blob/f77725bb168df62cd9c8db1f8f35a9d2a6ec88ae/src/mewbot/tools/paths.py)). 

Notably, at that point all are thin wrappers for python scripts, so having the workflow invoke the path ptyon script and then the test python script would work.

exachixkitsune:
Might be that there isn't an issue and it was user error? Thanks to help, it's identified that the newline characters in the git checkout were causing an issue.
That said I now still cannot duplicate the actual issue in the actions...

.github/workflows/review.yaml
```diff
@@ -77,15 +81,27 @@ jobs:
...
       - name: Upload Artifacts
         uses: actions/upload-artifact@v3
         if: always()
         with:
-          name: test-results-${{ matrix.version }}
+          name: test-results-${{ matrix.os }}-${{ matrix.version }}
           path: reports
 
+  tests-passed:
+    name: Tests Passes
```
javajawa:
```suggestion
    name: Tests Passed
```

javajawa:
> That said I now still cannot duplicate the actual issue in the actions...

Hmmm, might be worth dumping some debugging into ./tools/test on this branch?

exachixkitsune:
Current status:
If I run the tools/* scripts using bash such as `bash tools/test` it runs it in WSL, which works because WSL takes the path information from `tools/path` BUT it uses it's own, bespoke version of python and not the python from venv or windows.
 -> Which means the error I'm currently getting is that pytest isn't installed or similar, because it's not in the WSL instance.
 -> But potentially that can be isolated into a new issue, but solving that _may_ affect the solution to this.
This impinges my ability to test away from prod.

But that doesn't explain why it doesn't work on the actions, which should be using bash and thus should be working.

Tests to try:
1. Is the python path being set correctly when `tools/test` runs `tools/path`.
  -> if not, that's the way to solve it; refactoring path into a system that works.
  -> it it does, possible that a path modification needs to be appended into `mewbot.tools.test.py`?
2. Check path in `mewbot.tools.test.py`.
  -> Well, this shouldn't run because `mewbot.tools.test` isn't running...
3. Check to see what python thinks it's environment variables should be.
  -> I.e does it want PYTHONPATH or for some reason is it looking at something else?

Testing will occur?

exachixkitsune:
Ok so
PYTHONPATH in ubuntu wants `:` as the delimiter
PYTHONPATH in windows wants `;` as the delimiter
This can be found with `os.pathsep`.

As discussed, the fix has been to implement part of the planned update described [https://github.com/mewler/mewbot/pull/142#issuecomment-1444289197](in this comment), implementing a programatic way of configuring the path.

tools/path
```diff
@@ -7,5 +7,5 @@
 PYTHON=${PYTHON:-python3}
 readonly PYTHON
 
-PYTHONPATH="./src:./test"
-export PYTHONPATH
+PYTHONPATH="$("$PYTHON" src/mewbot/tools/path.py)"
+export PYTHONPATH
```
javajawa: Add EOL here

javajawa (commented):
Looking good!
3 linting warnings from Sonar, one for `tools/path` (see comment), and if you wouldn't mind fixing the typo I highlighted earlier in the thread.

PR #145, opened 2023-03-03
api-v1/Behaviour: Relax restrictions on Components

javajawa:
v1 Behaviours no longer require that all Components added to them are from the v1 base class, only that they conform to the Component protocols. This will allow Components from other ancestries to be used. However, it does mean that Behaviours are no longer guaranteed to be serialisable by the `mewbot.loader` code.

Additionally, the logic for what components have been added has been changed. Previously, a condition or action could only be added if there was a Trigger that accepted a similar kind of input event. However, the inheritance rules of InputEvent can get complicated, and the restrictions that were implemented were not always correct.

Consider a `DiscordMessage` which extends `DiscordInputEvent` and also extends or implements a `ReplyableEvent` protocol. If you had a Trigger that looks for `DiscordInputEvent`s in a specific guild or channel, and an action which replies to any `ReplyableEvent`, the old code would say there is no connection between those two, as neither is a superclass of the other.

Instead, the Behaviour now attempts to keep track of the widest set of classes its Triggers will accept. This may be used by the runner to pre-filter which InputEvents are sent to the Behaviour.

### DIFF
--- a/src/mewbot/api/__init__.py
+++ b/src/mewbot/api/__init__.py
@@ -3,3 +3,10 @@
 # SPDX-FileCopyrightText: 2021 - 2023 Mewbot Developers <mewbot@quicksilver.london>
 #
 # SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Container module for MewBot development APIs.
+
+This module will never contain any content, but will instead have child modules
+for different API versions and common components.
+"""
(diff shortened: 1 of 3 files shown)


javajawa:
Marking as draft until I get around to writing the test cases

ajCameron (approved):

PR #146, opened 2023-03-03
tools: Add script for documentation generation

javajawa:
Currently, the only generated documentation is the package map SVG.


exachixkitsune:
Do we not, in some way, need to ask for "Graphviz" to be installed to make the graphs?
Source: https://pypi.org/project/pydeps/#how-to-install

Re-reading, noting now that's it's a separate program.

javajawa:
> Do we not, in some way, need to ask for "Graphviz" to be installed to make the graphs? Source: https://pypi.org/project/pydeps/#how-to-install
> 
> Re-reading, noting now that's it's a separate program.

Notably it's a separate non-Python program (which is why it's not handles as a transitive dependency).
I'm personally happy for the failure mode to be whatever pydeps does is `dot` / `graphviz` is not present (though I do not know what that is)

exachixkitsune:
> > Do we not, in some way, need to ask for "Graphviz" to be installed to make the graphs? Source: https://pypi.org/project/pydeps/#how-to-install
> > Re-reading, noting now that's it's a separate program.
> 
> Notably it's a separate non-Python program (which is why it's not handles as a transitive dependency). I'm personally happy for the failure mode to be whatever pydeps does is `dot` / `graphviz` is not present (though I do not know what that is)

It makes an empty .xml file, overwriting the previous.

exachixkitsune (approved):
Happy

PR #148, opened 2023-03-04
api/Registry: Add component module detection

javajawa:
Modules which provide the `mewbot-v1` entrypoint in their module declaration can now be automatically detected and loaded. This will allow, in the future, for systems to retrieve a list of components that are available for use.

Additionally, the documentation for the Registry has been improved

### DIFF
--- a/src/mewbot/api/registry.py
+++ b/src/mewbot/api/registry.py
@@ -4,8 +4,16 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-"""Tooling for recording the creation of implementation classes, allowing
-for lists """
+"""
+Automation for registering components and plugins with a MewBot runner.
+
+Plugins are loaded using the metadata from python packages.
+Any package that exposes a `mewbot-v1` entrypoint will be assumed to contain
+components that can be used.
+
+This module also contains the ABCMeta subclass 'Registry'. All classes that use
+this as their metaclass will be recorded, forming an index of all loaded components.
+"""
 
 from __future__ import annotations
 
@@ -14,12 +22,20 @@ from typing import List, Type, Any, Dict, Optional, Callable, Iterable, Tuple
 import abc
 import uuid
 
+import importlib_metadata
+
 from mewbot.core import ComponentKind, Component
+
+
+API_DISTRIBUTIONS = ["mewbot-v1"]
 
 
 # noinspection PyMethodParameters
 class ComponentRegistry(abc.ABCMeta):
-    """ComponentRegistry is a AbstractBaseClasses MetaClass which instruments
+    """
+    Metaclass for Registering Component Classes.
+
+    ComponentRegistry is a AbstractBaseClasses MetaClass which instruments
     class definition to automatically record and classify classes implementing
     MewBot's interfaces, and then modifies instance creation to assign properties.
 
@@ -44,13 +60,15 @@ class ComponentRegistry(abc.ABCMeta):
     _api_versions: Dict[ComponentKind, Dict[str, Type[Component]]] = {}
 
     def __new__(mcs, name: str, bases: Any, namespace: Any, **k: Any) -> Type[Any]:
+        """
+        Hook for creating a class in this ancestry.
+
+        We confirm that is only implements one API, and then register it for later use
+        """
+
         created_type: Type[Any] = super().__new__(mcs, name, bases, namespace, **k)
 
-        if created_type.__module__ == mcs.__module__:
-            return created_type
-
         api_bases = list(mcs._detect_api_versions(created_type))
-
         if len(api_bases) > 1:
             raise TypeError(
                 f"Class {created_type.__module__}.{created_type.__name__} inherits from two APIs"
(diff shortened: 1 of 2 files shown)


javajawa:
Those were meant to be one commit; will squash when I add the relevant tests.

src/mewbot/api/registry.py
```diff
@@ -121,14 +165,48 @@ def do_register(api: Type[Component]) -> Type[Component]:
...
+
+        This function will only return the first matched type, as per the Registry's
+        'one base type' constraint.
+        """
+
+        if not isinstance(component, type):
+            component = type(component)
+
+        for val in mcs._detect_api_versions(component):
             return val
```
ajCameron, line 188: Suggested change - as far as I can tell, for well designed code which is behaving how we'd prefer, this should never trip more than once (if nothing else, there's a check elsewhere which makes it unlikely).

Thus I propose it should error, instead of returning the first matched value, if there is more than one match. Seems less likely to produce unexpected behavior and a safer behavior. People could still work around it, but at least they would know they where doing something a bit didgy. 

(But there might well be some use case I'm missing which would be a shame to loose, so I'm not married to the idea).

ajCameron (dismissed):

ajCameron:
(Overall comment - code looks really good! Very elegant)

src/mewbot/api/registry.py
javajawa: I've written up a change (754854a) that makes this examine the list, but I'm wondering what situations a second iteration could happen excluding people fiddling with subclass hooks -- and I'm no longer sure its worth defending against that?

ajCameron: Hmm. Yeah - it _should_ never occur unless, as you say, someone has messed with subclass hooks.

It just seems like the sort of assumption that might trip someone up at some point (it should match once and only once, but, currently, it silently works if, afact, there's more than one match).

It's probably not an issue! I'd only suggest closing it off out of an abundance of caution.

exachixkitsune: My opinion would be if it's a potential error state - even if we don't know how the error could occur - it should report an error; but then that's just in the effort of robustness not efficiency.

src/mewbot/api/registry.py
```diff
@@ -121,14 +165,48 @@ def do_register(api: Type[Component]) -> Type[Component]:
...
 
         raise ValueError(f"No API version for {component}")
+
+    @staticmethod
+    def load_and_register_modules(name: Optional[str] = None) -> Iterable[Any]:
+        """
+        Load modules from setuptools that declare a `mewbot-v1` entrypoint.
```
exachixkitsune: Does this comment want to say "mewbot-v1" or "setuptools that declare an entrypoint matching the current api"
```suggestion
        Load modules from setuptools that declare a entrypoint with a valid API.
```

src/mewbot/api/registry.py
```diff
@@ -121,14 +165,48 @@ def do_register(api: Type[Component]) -> Type[Component]:
...
+        This looks at the entrypoints of all modules which are locatable with
+        importlib,
+
+        :param str name: if given, loads only plugins with the given `name`.
+        :return: the loaded objects.
+        """
+        for distribution in importlib_metadata.distributions():  # type: ignore
+            for entry_point in distribution.entry_points:
+                if entry_point.group not in API_DISTRIBUTIONS:
+                    continue
```
exachixkitsune, line 212: On the thought of error dectection, is it worth saying "This item's not being picked up because of this reason"? so we're not like "Why is my module not loading? :("

exachixkitsune (dismissed):
I think I approve of a bunch of this.

src/mewbot/api/registry.py
javajawa: That message would occur for every python module installed in the current context, which I don't think would be productive for anyone?

src/mewbot/api/registry.py
javajawa: Hmm, interesting. If we're going for that, I'd probably say `supported`?

src/mewbot/api/registry.py
exachixkitsune: ah. that's fair yeah maybe let's not do that. Is there some other confirmation of what modules load?

src/mewbot/api/registry.py
exachixkitsune: Sounds reasonable to me
```suggestion
        Load modules from setuptools that declare an entrypoint with a supported API.
```

src/mewbot/api/registry.py
javajawa: I'm not quite sure what you mean by "confirmation"? This isn't intended as an interactive process, and even having a name filter is mostly for having the same signature as `pluggy`

exachixkitsune: Resolved via direct discussion

src/mewbot/api/registry.py
javajawa: 754854a has been included in the tree

exachixkitsune (approved):
Seems ok to me!

PR #149, opened 2023-03-04
tools: Configure to work with any src-dir projects

javajawa:
This change is the first step for introducing the concepts of both `mebot.tools` as an external package which can be pulled in to lint another codebase (such as a mewbot plugin), and for restructuring the repository to have in-tree plugins.

The change adds the capability for `mewbot.tools.path` to dynamically generate paths by looking for folders matching a certain name in a given location. Currently it looks only in `.`, but this can easily be expanded.

Along with these changes, the ability to filter what paths are examined by the paths has been added, with both the `lint` and `test` tool chains taking a list of paths to run against, and the `test` tools also supporting a list of coverage references being selected.

This allows for easier checking of the state of a specific module. For example

```sh
./tools/test tests/io/test_io_rss.py --cover mewbot.io.rss
```

Will run only one test file, and produce a coverage report for a single class. This should reduce testing times when working locally on a specific component.

Additionally, the entire module's documentation has been updated.

### DIFF
--- a/src/mewbot/tools/path.py
+++ b/src/mewbot/tools/path.py
@@ -4,14 +4,43 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-from typing import List
+"""
+Tools for detecting paths at which source files are installed.
+
+This file, when called directly, prints out a set of values for PYTHONPATH.
+As such, it can not rely on anything that would be loaded using PYTHONPATH,
+i.e. we can not use any other mewbot code from this file.
+"""
+
+from collections.abc import Iterable
 
 import os
+import pathlib
 
 
-def gather_paths() -> List[str]:
-    return ["./src", "./test"]
+def scan_paths(
+    root: pathlib.Path, *filters: str, recursive: bool = True
+) -> Iterable[pathlib.Path]:
+    """Scan for folders with a given name in the provided path."""
+
+    if not recursive:
+        yield from [root / name for name in filters if (root / name).exists()]
+        return
+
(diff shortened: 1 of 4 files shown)


javajawa:
Interestingly, Sonar doesn't like me replacing `typing.Generator[X, None, None]` with `collections.abc.Iterable[X]`, but I think that should be a valid replacement due to https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes and https://peps.python.org/pep-0585/.

It's still passing the quality gate, so it's non-blocking for this change, but we might need to upstream and issue.

src/mewbot/tools/path.py
```diff
@@ -4,14 +4,43 @@
...
 
-def gather_paths() -> List[str]:
-    return ["./src", "./test"]
+    return (str(x.absolute()) for x in scan_paths(root, *filters, recursive=False))
 
 
 if __name__ == "__main__":
-    print(os.pathsep.join(gather_paths()))
+    # When called directly, this module output the value of the PYPATH environment variable
+    print(os.pathsep.join(gather_paths("src")))
```
exachixkitsune, line 46: Is the removal of `./test` here deliberate?

exachixkitsune (approved):

src/mewbot/tools/path.py
javajawa: Yes -- `./test` isn't a proper package hierarchy, except for the old `tests.common` package which has been migrated to `mewbot.test`

PR #152, opened 2023-03-04
actions: Only lint merge files that have changed

javajawa:
When running in a merge request context, we only need to run the linters against the changed files. This aligns better with the Sonar concept of not penalising merge requests for code they have not changed, and will allow for easier enabling of new linting tools (such as documentation requirements).

### DIFF
--- a/.github/workflows/review.yaml
+++ b/.github/workflows/review.yaml
@@ -28,6 +28,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 100
 
       - name: Setup Python ${{ env.DEFAULT_PYTHON }}
         uses: actions/setup-python@v4
@@ -42,7 +44,13 @@ jobs:
         run:  pip install -r requirements-dev.txt
 
       - name: Run Linters
-        run: ./tools/lint
+        run: |
+          if [ -n "${{ github.event.pull_request.base.sha }}" ]
+          then
+            git diff --name-only --diff-filter=ACMRT "${{ github.event.pull_request.base.sha }}" "${{ github.sha }}" | grep '.py$' | xargs ./tools/lint
+          else
+            ./tools/lint
+          fi
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v3


javajawa:
@exachixkitsune @ajCameron I'm...not 100% convinced by my own arguments that this is a good idea.
It does solve the problem as noted in #151 of setting up documentation linting when a lot of our code is missing documentation, but some tools like `pylint` use context of the entire codebase to output information -- selecting specific files might degrade those tools outputs.

exachixkitsune:
The question may be "how often does linting partially (i.e. only changed files) cause an issue". I'd suggest maybe it could; particularly mypy checking types (if only limited to changed files, will it see where code is used and where there may be a typing issue?), or pylint, for whom the description is "pylint performs a similar set of checks as flake8, but does so using the full codebase as context".

It's also a mask for the issues in #151, which is only tolerable due to the low maturity of the codebase (i.e., not in production rather than otherwise)

Even if this is implemented, I presume it would still run in it's entirety at some point? I would suggest at least that at some point.

javajawa:
`mypy` will still inspect types from modules not named on the command line (that's how it pulls in external typing).

I suspect, but have not verified, that `pylint` will (for example) not detect duplicated code if one of the copies falls outside of the scanned code.

This is just for the merge request variation, i.e. where there is a distinct set of changes. For branches and tags the full lint will still happen.

ajCameron:
My opinion is we should continue to lint the whole code base at every stage.

My reasons are
1) It's simply easier - though _probably_ only slightly. As an extension to this, it's conceptually cleaner
2) Nagging from linters may encourage people to fix other parts of the code base that need it - raising `pylint` code quality is a generally good thing, and something I am encouraged to do when I see the annoying error messages. Even in parts of the code I am not currently working on. Thus global linting has produced at least some tangiable benefits. At least from my side (if only  a matter of motivation).
3) I agree with @exachixkitsune. It's a mask for issues we should probably just attempt to solve as fast as possible. Before addressing them becomes even more time prohibitive.
4) I believe continuing to use `pylint` for global code duplication detection - beyond the current changed files - is a highly useful feature. It would be a real shame not to have it anymore.

I believe the current outstanding linting issue is largely over doc strings.

Thus, I propose that we get the current series of pull requests on the mainline (I will have a look over them today and tomorrow). Then do a big PR to add doc strings and other documentation.

ajCameron:
(Can reassess in future if we end up wanting to enable additional linting features)

javajawa:
I will, after #151 is merged, raise a new merge that enables `pydocstyle` for changed files as a check so that we can ensure that new code is correctly documented, until we reach the point that we can finish out the documentation.

PR #153, opened 2023-03-05
core: Add and update pydocs

javajawa:
This updates the pydocs for the `mewbot.core` module. Mostly these are lifted from #145, with some changes to remove v1 specific features, and also explain broader API contracts.


ajCameron (approved):
Good docs! Clear and thoughtful.

PR #155, opened 2023-03-06
Enabling all missing docstring warnings in pylint and getting code base to pass pydocstyle

ajCameron:


ajCameron:
I'm going through pylint, adding back in docstring types one by one.

When this is complete (probably tommorrow) I'll do a pass to fix all pydocstring complaints which are not concerning docstring existence.

src/mewbot/demo.py
```diff
@@ -2,6 +2,12 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
+"""
+Simple example elements which demonstrate some of the most basic mewbot objects.
```
javajawa, line 6: I think this file might be redundant give the examples system?

src/mewbot/io/desktop_notification.py
```diff
@@ -4,6 +4,10 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
+"""
+Provides an IOConfig with the capability to produce desktop notifications on (most) systems.
```
javajawa:
```suggestion
Provides an IOConfig with the capability to produce desktop notifications.

Notification support is dependent on the system running the bot supporting
standardised notification interfaces. Current support includes Windows 10,
and any Linux build with `notify-send` installed.
```

src/mewbot/io/discord.py
```diff
@@ -168,6 +207,12 @@ async def run(self) -> None:
 
 
 class InternalMewbotDiscordClient(discord.Client):
+    """
+    Discord.Client with overrode methods to actually interact with mewbot.
+
+    (In particular, methods have been overridden to write events to the InputQueue when they occur)
```
javajawa:
```suggestion
    In particular, methods have been overridden to write events to the InputQueue when they occur
```

src/mewbot/io/rss.py
```diff
@@ -83,6 +90,12 @@ class RSSInputEvent(InputEvent):
 
 @dataclasses.dataclass
 class RSSOutputEvent(OutputEvent):
+    """
+    NOT CURRENTLY SUPPORTED.
+
```
javajawa: I'd honestly suggest cutting this class for now.

src/mewbot/api/registry.py
```diff
@@ -132,6 +134,12 @@ def register_api_version(
         """
 
         def do_register(api: type[Component]) -> type[Component]:
+            """
+            See the doc string of the function which returns this function.
```
javajawa: Feels that

> See :mewbot.api.registry:register_api_version

would be more user friendly. (but in whatever the sphix syntax is.

src/mewbot/io/socket.py
```diff
@@ -30,11 +44,20 @@ class SocketIO(IOConfig):
     _socket: Optional[SocketInput]
 
     def __init__(self) -> None:
+        """
+        Initialise the SokcetIO IOConfig.
```
javajawa:
```suggestion
        Initialise the SocketIO IOConfig.
```

src/mewbot/loader.py
```diff
@@ -4,6 +4,12 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
+"""
+Provides functions to load mewbot components from YAML.
+
+All functions which load components from YAML should go here.
```
javajawa: minor: not sure this last sentence adds anything?

tests/test_core.py
```diff
@@ -15,56 +21,101 @@
...
 class TestComponent:
-    # Test a set of what interface passing succeeds and fails
+    """
+    Test a set of what interface passing succeeds and fails.
+    """
+
     @staticmethod
     def test_componentkind_interface_map_behaviour() -> None:
+        """
+        The interface method returns the interface supported by that Component.
```
javajawa, line 38: These are missing line-breaks after the first sentence.
_However_, pydocstyle ignores tests by default, and I have already addressed this in 2dd84440c2954cae5f0c7c90d9751a21f254e4de which can be a future merge

javajawa (commented):
Looking very good overall.

There's a a few classes and files which I feel might be better served just being deleted than documented, and a couple of minor documentation thoughts.

src/mewbot/demo.py
ajCameron: Currently the basic elements here are used in a few places in the yaml example files.

I think there's some value to keeping some basic elements - stuff useful for debug - in the main code and not in the examples (which we might end up not shipping in all cases). Suggest shifting them to something like mewbot.utils.common_elements?

javajawa: I suspect there's going to have to be an API module for Event protocols (e.g. `Replyable`) and non specific conditions and actions so it might be worth taking the actual components from this module over there, and the rest out to examples?

ajCameron: Makes sense to me! Do you want to part it down? (as I'm not sure where you'd want it to go in the API module).

If you do, I can take a look this evening.

javajawa: Moved the thought to https://github.com/mewler/mewbot/issues/162 as it's not a part of this change and isn't pressing.

javajawa (approved):

Issue #143, opened 2023-02-25
Tools to be rendered able to be ran on Windows


exachixkitsune:
As found in PR#142, tools don't run easily in windows, they take some work to get to work.

This issue is to get those tools to work.

ajCameron:
I've got a plan - make python versions of them all and allow them to be called directly as a module through mewbot.tools.

I'll get on it sometime this weekend or early next week.

javajawa:
Currently, all of them have python versions except `tools/docs` (which I suspect will get picked up in #161) which can be invoked as `python -m mewbot.tools.[tool]` assuming the `PYTHONPATH` is currently correct. I think maybe dusting off #49  and configuring the `.pth` from `mewbot.tools.paths` will solve most of the technical problem, and then updating the docs to refer to the python module names by default and mentioning somewhere in the README that the shorter scripts are available on some systems?

PR #160, opened 2023-03-08
Setup in-tree plugin handling

javajawa:
MewBot plugins are generally intended to live in their own repositories
and have their own release cycles. However, some small and core plugins
may want to live "in-tree" in this repository.

For clarity, these can now be moved to
  `plugins/[package-name]/{src,tests}`

As these plugins both use and provide the `mewbot.io.*` packages,
`mewbot` and `mewbot.io` have been changed into "namespace packages"

https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

A new tool has been added to simplify installing dependencies.
`./tools/install-deps` / `mewbot.tools.install` re-uses the
`gather_paths` to look for `requirements.txt`.

### Update path detection for MyPy

MyPy does not use the stock import engine for doing its analysis,
so we have to give it additional hints about how the namespace package
structure works.
See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules

There are two steps to this:
 - We set MYPYPATH equivalent to PYTHONPATH
 - We alter the folder list such that, in src-dir folders, we pass the
   folder of the actual pacakge (i.e. ./src/mewbot rather than ./src)


javajawa:
I have added onto an existing Sonar bug: re the Iterable/Generator confusion
https://community.sonarsource.com/t/python-s5886-false-positive-for-generators-annotated-with-iterator-for-return-type/79615/9

.github/workflows/review.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
             requirements-dev.txt
```
exachixkitsune, line 75: Is this prior line going to be affected by the dynamic generation from `install-deps`?

plugins/mewbot-io-discord/src/examples/discord_bots/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+Classes for providing examples for mewbot-io-discord.
+"""
```
exachixkitsune:
```suggestion
"""

```

src/mewbot/tools/lint.py
```diff
@@ -90,13 +93,30 @@ def lint_mypy(self) -> Iterable[Annotation]:
...
             args.append("--pretty")
 
+        # MyPy does not use the stock import engine for doing its analysis,
+        # so we have to give it additional hints about how the namespace package
+        # structure works.
+        # See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules
+        #
+        # There are two steps to this:
+        #  - We set MYPYPATH equivalent to PYTHONPATH
+        os.putenv("MYPYPATH", os.pathsep.join(gather_paths("src")))
```
exachixkitsune, line 109: is it worth having a single func that reads "return_os_path(in)"; such as:
```suggestion
        os.putenv("MYPYPATH", return_os_path("src"))
```

exachixkitsune: Adding the rest of this code is A Lot for a suggestion, but I can do to a commit if desired

exachixkitsune (dismissed):
Seems good to me, though I'd reccomend #163 implemented first in order to ensure docs are all good

.github/workflows/review.yaml
javajawa: Technically it will, but that's frankly very hard to actually program.
The cache will still be useful -- it'll get read based on the two main files, and the downloaded packages from the install step will get written back, but it will take an efficiency hit when we change a dependency in a file not listed here.

Might be worth an issue?

src/mewbot/tools/lint.py
javajawa: Would that function be anything more than
```python
def gather_os_path_str(*folders: str) -> str:
    return os.pathsep.join(gather_paths(*folders))
```

javajawa:
So, the failing mypy test...runs fine locally. That's going to fun to debug.

exachixkitsune:
> So, the failing mypy test...runs fine locally. That's going to fun to debug.

Good news/bad news; I am able to duplicate the mypy error; except it's erroring on a different module each time...

exachixkitsune:
Addendum; I am using mypy-1.1.1, alongside github.
feels like it's https://github.com/python/mypy/issues/8944 occuring again, but we're also already engaging the fix.
Enforcing mypy-0.991 didn't fix it

javajawa:
I get successes with both of the following builds:
```
$ pip list | grep -e mypy -e typing
mypy                                    1.0.1
mypy-extensions                         1.0.0
typing_extensions                       4.4.0
```

```
$ pip list | grep -e mypy -e typing
mypy                                    1.1.1
mypy-extensions                         1.0.0
typing_extensions                       4.4.0
```

which seems to imply it's more something deeper environmental issue...

Interestingly, if I checkout a new copy of the code and run the test, I also encounter the same issue, so it's something exceptionally specific.

javajawa:
Ok, I have a reliable reproducer for both the broken and working states: the presence of `__pycache__`.

```
$ git status --ignored                                                                                        
On branch plugin-paths
Your branch is up to date with 'origin/plugin-paths'.

Ignored files:
  (use "git add -f <file>..." to include in what will be committed)
        .coverage
        .idea/
        .pytest_cache/
        TODO
        reports/
        src/mewbot/__pycache__/
        src/mewbot/api/__pycache__/
        src/mewbot/core/__pycache__/
        src/mewbot/io/__pycache__/
        src/mewbot/test/__pycache__/
        src/mewbot/tools/__pycache__/
        tests/__pycache__/
        tests/api/__pycache__/
        tests/io/__pycache__/
        venv/
```

```
$ ./tools/lint                    
All done! ✨ 🍰 ✨
35 files left unchanged.
There are no .py[i] files in directory '/home/benedict/src/javajawa/mewbot/tests/__pycache__'

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

```
$ find . -name __pycache__ -exec rm -R '{}' \+
```
```
$ ./tools/lint                                
All done! ✨ 🍰 ✨
35 files left unchanged.
tests/api/test_v1_behaviour.py: error: Duplicate module named "tests.api.test_v1_behaviour" (also at "/home/benedict/src/javajawa/mewbot/tests/api/test_v1_behaviour.py")
tests/api/test_v1_behaviour.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
tests/api/test_v1_behaviour.py: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
```

I restored the `__pycache__` folders using `./tools/test` and then mypy passes again.


To quote a knife missile in Hydrogen Sonata: "Actions to defeat/disable unclear"

src/mewbot/tools/lint.py
```diff
@@ -218,6 +238,27 @@ def lint_black_diffs(
...
+
+    "src-dir" style folders will be expanded out into the paths for the
+    root of each module. src-dir style roots are detected on the basis
+    of being members of `sys.path`.
+    """
+
+    for path in folders:
+        if path not in sys.path:
+            yield path
+            continue
```
javajawa: For the record, the issues were being caused by a combination of this `continue` being missing and that that resulted in the `__pycache__` being included in the include path, if it was already present.

If could also possible be rewritten as
```python
    for path in folders:
        if path in sys.path:
            potenital_paths = [os.path.join(path, module) for module in os.listdir(path)]
            yield from [module_path for module_path in potenital_paths if os.path.isdir(module_path)]
        else:
            yield path
```
which might be clearer?

src/mewbot/tools/lint.py
javajawa: Discussed elsewhere, doesn't seem particularly worthwhile where there would only be two callees.

src/mewbot/tools/lint.py
ajCameron: I think it is, but that could also be done via comments. Dealer's choice, really!

.github/workflows/review.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
...
 
       - name: Install dependencies
-        run:  pip install -r requirements-dev.txt
+        run: ./tools/install-deps
```
ajCameron, line 42: Note - this is not currently working on windows - see pull request https://github.com/mewler/mewbot/pull/49 - in particular the additions to tools/path.sh.

As these are windows only fixes, seemed easier to include them in the general windows docs. But I can pull them out and make em a seperate request if you'd prefer for neatness

src/mewbot/tools/install_deps.py
```diff
@@ -0,0 +1,36 @@
...
+
+    file: pathlib.Path | str
+    requirements: list[str] = []
+
+    for file in dot.glob("requirements-*.txt"):
+        requirements.extend(("-r", str(file)))
+
+    for file in gather_paths("requirements.txt"):
+        requirements.extend(("-r", str(file)))
+
```
ajCameron, line 30: Seems a good way to do this

ajCameron (dismissed):
Looks good! See comment about the hack to tools/path.sh - but we can deal with that later.

src/mewbot/tools/lint.py
```diff
@@ -218,6 +238,27 @@ def lint_black_diffs(
...
+
+        yield from [
+            module_path
+            for module_path in [os.path.join(path, module) for module in os.listdir(path)]
+            if os.path.isdir(module_path)
+        ]
```
javajawa:
```suggestion
    for path in folders:
        if path in sys.path:
            # Each folder in PYTHONPATH is considered a src-dir style collection of modules
            # We xpand these into the actual list of modules
            potential_paths = [os.path.join(path, module) for module in os.listdir(path)]
            yield from [module_path for module_path in potential_paths if os.path.isdir(module_path)]
        else:
            # All other paths are returned unchanged
            yield path
```

.github/workflows/review.yaml
javajawa: We've got #143 which is pointing to #49 -- I think that's good enough?
All the tools should work even on Windows given a posix-ish shell to run them in, otherwise we have the option to directly invoke the underlying python.

(Maybe we can open an issue with MicroSoft to have a posix shell on all windows machines by default?)

ajCameron: I was wondering about adding direct invoke to the docs - if we're committed to having the shell scripts just be a tight wrapper around the underlying python invoke (which makes sense to me) - I'll go ahead and add it in as an alternative to installing a posix shell (which is a bit of a faff) to https://github.com/mewler/mewbot/pull/49

exachixkitsune (approved):
As I feel like my comments have been addressed, am happy.

ajCameron (approved):
Looks good!

PR #161, opened 2023-03-10
Adding sphinx - first builds of sphinx docs

ajCameron:
The purpose of this request is to gather comment on
 - how the docs appear (theme choice e.t.c)
 - any additional information that should be automatically included from the code base (currently consisting of a full module guide generated from the modules' docstrings)

Currently the docs creation toolchain is not present - each step needs to be taken manually. I thought it would be useful to get some comments on the current form of the docs in general before I worked on that problem (as we might want to add some custom css, stuff like that, and knowing that before I go diving into the sphinx internals would be good).

For details of the proposed toolchain, see docs/README_DOCS.md. In particular the section "How do I build this documentation?". It is my intent to build a toolchain to hook into the sphinx internals and automate this entire process.

For the docs themselves, please start at docs/build/html/index.html.

All comments welcome!


javajawa:
Initial thought: we should not commit the built docs to this repo:
 - it means that every merge request would require a doc rebuild, which would have to be verified
 - it's _heavy_ -- this is considerably larger than our code
 - developer's can build the docs if they need them

For user's who want to read the documentation, I would suggest instead:
 - create a new repository `mewbot-docs`
 - set up an action on this repository that, on push to mainline or a release happening, [dispatches](https://github.com/marketplace/actions/repository-dispatch) an event to mewbot-docs
 - the event runs an action that grabs the commit (mainline or tag), builds the docs, and commits it to it's own mainline (or equivalent tag)

This gives us a repo of the docs, which is self-updating and (if we design it correctly) multiple sources can push into (so "official" plugins could be documented at the same time). We can even publish that repo with github-pages!

ajCameron:
Excellent notion!

This would also allow cutting down requirements-dev.txt - can shift the docs _build_ (but not validation) stuff into requirements-docs.txt

Will convert this to draft and get it set up that way.

Issue #162, opened 2023-03-11
Deprecate and split `mewbot.demo`


javajawa:
The "demo" module is a combination of useful components and non-reusable examples that were written back at the beginning of the project.

This should be moved -- see the discussion in https://github.com/mewler/mewbot/pull/155#discussion_r1132156689

javajawa:
This has been added to #166

PR #166, opened 2023-03-18
Refactor output queue

javajawa:
## Refactor Actions to not require knowledge of queues
    
This change makes each call to `Behaviour.process` and `Action.act`
less stateful by moving all references to the queue out to the bot
or runner levels of interaction.
Both of these methods now emit (in the form of an async generator)
all output events that would be created.

As many actions will not emit any events (for example for updating
state information of recording data to datastores), the ability to
`yield None` has been codified in. This allows dvelopers to ensure
they are providing an async generator.

## Refactor mewbot.demo into mewbot.io.common
    
The demo module was written very early in the project's development,
and includes code that doesn't do anything. The use pieces of code
have been moved to a new IO module, with tests added.

This resolves issue #166.

## Add 'Reply' generic events and action to the Common IO.
    
A InputEvent abstract class/mix-in/protocol has been added which
allows an InputEvent to create an OutputEvent which is a reply to it.

This event contains a string message, to which no formatting information
is assigned.

There are two scopes for the reply -- either the same scope as the
incoming message, or in a potentially 'narrower' scope which still
includes the sender, but fewer other people (such as a direct message).

These two modes may be equivalent.

A ReplyAction has also been added, using the `string.Template` to
format the current state into the message, making it able to combine
arbitrary IO systems with other arbitrary actions.

### DIFF
--- a/examples/rss_input.yaml
+++ b/examples/rss_input.yaml
@@ -18,11 +18,11 @@ properties:
   name: 'RSS Printer'
 triggers:
   - kind: Trigger
-    implementation: mewbot.demo.AllEventTrigger
+    implementation: mewbot.io.common.AllEventTrigger
     uuid: aaaaaaaa-aaaa-4aaa-0000-aaaaaaaaaa03
     properties: { }
   - kind: Trigger
-    implementation: mewbot.demo.AllEventTrigger
+    implementation: mewbot.io.common.AllEventTrigger
     uuid: aaaaaaaa-aaaa-4aaa-0000-aaaaaaaaaa04
     properties: { }
 conditions: []
--- a/examples/trivial_http_post.yaml
+++ b/examples/trivial_http_post.yaml
@@ -18,12 +18,12 @@ properties:
   name: 'Echo Inputs'
 triggers:
   - kind: Trigger
-    implementation: mewbot.demo.AllEventTrigger
+    implementation: mewbot.io.common.AllEventTrigger
     uuid: aaaaaaaa-aaaa-4aaa-0002-aaaaaaaaaa02
     properties: { }
 conditions: []
 actions:
   - kind: Action
-    implementation: mewbot.demo.PrintAction
+    implementation: mewbot.io.common.PrintAction
     uuid: aaaaaaaa-aaaa-4aaa-0002-aaaaaaaaaa03
     properties: { }
--- a/examples/trivial_socket.yaml
+++ b/examples/trivial_socket.yaml
@@ -18,12 +18,12 @@ properties:
   name: 'Echo Inputs'
 triggers:
   - kind: Trigger
-    implementation: mewbot.demo.AllEventTrigger
+    implementation: mewbot.io.common.AllEventTrigger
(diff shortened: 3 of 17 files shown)


javajawa:
@ajCameron @exachixkitsune Ok, so this "works".

I did try to get it so `Action`s could just return a list rather than yielding, so that all of the following would be acceptable.
```python
async def act(...) -> AsyncIterable[OutputEvent] | Iterable[OutputEvent] | None:
    yield OutputEvent()

async def act(...) -> AsyncIterable[OutputEvent] | Iterable[OutputEvent] | None:
    return []

async def act(...) -> AsyncIterable[OutputEvent] | Iterable[OutputEvent] | None:
    print("Hello")
```

However, the typing got...really weird.

```
tests/api/test_v1_behaviour.py:158: error: Argument 1 to "add" of "Behaviour" has incompatible type "TestAction"; expected "Union[TriggerInterface, ConditionInterface, ActionInterface]"  [arg-type]
            behaviour.add(action)
                          ^~~~~~
tests/api/test_v1_behaviour.py:158: note: "TestAction" is missing following "TriggerInterface" protocol member:
tests/api/test_v1_behaviour.py:158: note:     matches
tests/api/test_v1_behaviour.py:158: note: "TestAction" is missing following "ConditionInterface" protocol member:
tests/api/test_v1_behaviour.py:158: note:     allows
tests/api/test_v1_behaviour.py:158: note: Following member(s) of "TestAction" have conflicts:
tests/api/test_v1_behaviour.py:158: note:     Expected:
tests/api/test_v1_behaviour.py:158: note:         @overload
tests/api/test_v1_behaviour.py:158: note:         def act(self, event: InputEvent, state: Dict[str, Any]) -> AsyncIterable[OutputEvent]
tests/api/test_v1_behaviour.py:158: note:         @overload
tests/api/test_v1_behaviour.py:158: note:         def act(self, event: InputEvent, state: Dict[str, Any]) -> Coroutine[Any, Any, Optional[Iterable[OutputEvent]]]
tests/api/test_v1_behaviour.py:158: note:     Got:
tests/api/test_v1_behaviour.py:158: note:         def act(self, event: InputEvent, state: Dict[str, Any]) -> AsyncIterable[OutputEvent]
```

so I have opted to go for just always forcing an aysnc generator.

There's a few possibilities, such as having two protocols, or two methods on the protocol, or...whatever. But I'm not sure it's worth the effort of tricking the typing system out of this state. There might also be an argument in favour of opening a ticket with the typing peeps to see if they have a suggestion.

ajCameron:
I mean, that's clearly not expected - or desirable - behavior. So I'd say a ticket makes sense.

javajawa:
Ok, some amount of scope creep happened here.

In order to get Sonar passing, I realised I was going to have to add tests to some code I had touched but not actually changed. This included the code to be removed in #162, so I have removed `mewbot.demo` and replaced the `Foo` condition with another class that has properties but actually serves a practical purpose.

ajCameron (approved):
Looks good!

Nice to bump test coverage in general. Shame it was because sonarcloud was being a pain.

PR #169, opened 2023-03-20
Doc builder prototype

ajCameron:
Removed sphinx output from commit - rationalized document builder script.

Docs are currently generated in docs_build.

Ready for comments on appearence and function - before setting up automated document build integration with git.


requirements-dev.txt
```diff
@@ -8,20 +8,23 @@
 -r requirements-test.txt
 
 # Installation toolchain
-setuptools~=67.6
+setuptools~=67.4
```
exachixkitsune: Why this reversion?

exachixkitsune (dismissed):
I am happy but I haven't tested it.

README.md
```diff
@@ -7,22 +7,26 @@ SPDX-License-Identifier: CC-BY-4.0
 # MewBot
 
 [![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](CODE_OF_CONDUCT.md)
+
```
javajawa, line 10: Adding these breaks in does mean that the widgets end up not in a row: [before](https://github.com/mewler/mewbot/tree/dcc0c549503a6d6c08815f71e2998a05149f14a9#mewbot) | [after](https://github.com/mewler/mewbot/tree/7c23024d1fdecf433bd335df735df7aaa69b796d#mewbot).

It is only now that I have noticed that ther all contributors comments are already doing that for that one badge. I'd prefer them in a line but this does make it look less weird than it currently does. So I'm happy to leave this change and sort it out another time.

docs/Makefile
```diff
@@ -0,0 +1,20 @@
...
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+%: Makefile
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
javajawa: this looks like it could be a string in `src/mewbot/tools`, save adding a new dependency on `make`

docs/README_DOCS.md
```diff
@@ -0,0 +1,127 @@
...
+1) Make sure you've run `pip install -r requirements-dev.txt` sometime recently.
+2) make sure you have your mewbot `venv` activated
+3) Run `python -m mewbot.tools.docs` to build
+
+# How to I contribute to these docs?
+
+More documentation is always welcome!
+These docs are intended as a living document.
+Feel free to add to them.
+In particular, details of problems you encountered - and how you solved them - would be most welcome.
```
javajawa: Probably worth linking back to contributing?

docs/README_DOCS.md
```diff
@@ -0,0 +1,127 @@
...
+9) Delete `_build` - to actually make these changes show up anywhere.
+10) Run `make html` (or any of the other output methods supported by `sphinx` - for a full list see their documentation)
+11) The doc files will be generated in the `_build/html` in this directory
+12) Open `_build\html\index.html` with your favourite web browser, and enjoy!
+
+## If you want to use mewbot's internal tooling (RECOMMENDED)
+
+1) Make sure you've run `pip install -r requirements-dev.txt` sometime recently.
+2) make sure you have your mewbot `venv` activated
+3) Run `python -m mewbot.tools.docs` to build
```
javajawa: The recommended method should come before the non recommended method, because humans

docs/README_DOCS.md
```diff
@@ -0,0 +1,127 @@
+
+# Welcome to the mewbot docs!
+
+Mewbot is a plugin friendly, multi-platform, chatbot framework which can load bots from YAML definitions.
+
+The docs are built using `sphinx`.
```
javajawa:
```suggestion
The docs are built using [`sphinx`](https://www.sphinx-doc.org/).
```

docs/README_DOCS.md
javajawa: Additionally, are we setting an explicit preference to `.rst` files that should be mentioned here? It's sort of implied by the "why is this a mix?" question.

docs/README_DOCS.md
```diff
@@ -0,0 +1,127 @@
...
+In particular, details of problems you encountered - and how you solved them - would be most welcome.
+
+# Doc strings
+
+Another very helpful thing you can do is add to, or further build out, the doc strings at module, class and function level.
+Sphinx uses the reStructuredText Docstring Format.
+
+See https://peps.python.org/pep-0287/ for more information.
+
+For the use of rst with sphinx, please see  https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html.
```
javajawa: I think most markdown renderers will convert these automatically, but I might still consider
```suggestion
See [PEP-0287](https://peps.python.org/pep-0287/) for more information.

For the use of rst with sphinx, please see the [Sphinx RTD tutorial section on docstrings](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html).
```

Not the biggest deal.

docs/conf.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+"""
+Configuration file for mewbot sphinx document build.
+"""
```
javajawa: The comments here feel like they should be part of the doc block.

docs/conf.py
```diff
@@ -0,0 +1,48 @@
...
+
+project = 'mewbot'
+copyright = '2023, Mewler Org'
+author = 'Mewler Org'
+release = '0.0.1'
```
javajawa, line 16: To raise as a separate issue: need to work out the source of truth for these values.

docs/conf.py
```diff
@@ -0,0 +1,48 @@
...
+
+html_theme = 'alabaster'
+html_static_path = ['_static']
+
+# The name of the Pygments (syntax highlighting) style to use.
+pygments_style = 'sphinx'
+
+
+
+
```
javajawa:
```suggestion
```

setup.py
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
 
 """
-Custom setup.py to install some, or all, mewbot pckages.
-
```
javajawa: I think changes in this `setup.py` all want to be reverted.

docs/make.bat
```diff
@@ -0,0 +1,35 @@
...
+if "%1" == "" goto help
+
+%SPHINXBUILD% -M %1 %SOURCEDIR% %BUILDDIR% %SPHINXOPTS% %O%
+goto end
+
+:help
+%SPHINXBUILD% -M help %SOURCEDIR% %BUILDDIR% %SPHINXOPTS% %O%
+
+:end
+popd
```
javajawa: I'm guessing this is a bat file implementation of the makefile?

javajawa (commented):
Initial code review pass.

Will comment on the changes to the ToolChain stuff when I test building the docs.

docs/Makefile
javajawa: Update: I have spotted `mewbot.tools.docs` which probably *is* that already. Is this file still useful?

requirements-dev.txt
ajCameron: I have _absolutely_ no idea. Corrected!

(Seems fine on a reinstall - so assuming it was an error. Thanks for the catch!)

README.md
ajCameron: Fair enough! I'd assumed this was a typo and the intent was a column - as there seemed to be a mix before.

docs/Makefile
ajCameron: It is not.

The thought was to leave it in in case someone wanted to use a Make integration - but there's no reason to encourage people to bypass the canonically preferable method (build through `mewbot.tools.docs`) - removing.

docs/README_DOCS.md
ajCameron: Link added.

... I was hoping not to (allowing people to use whatever format they feel more comfortable with), but the translation layer used to incorporate .md into sphinx is a lot less than perfect. So it's probbly worth stating the preference.

docs/README_DOCS.md
ajCameron: Yuuuuuup.

(Fixed)

docs/README_DOCS.md
ajCameron: Added

docs/conf.py
ajCameron: Agreed - corrected

docs/conf.py
ajCameron: Correcting.

docs/make.bat
ajCameron: It is - has been removed - as the canonical way to build is through the `.sh` script.

setup.py
ajCameron: Agreed. Corrected.

docs/conf.py
ajCameron: Ohh - yes. Good point. I'll add it as an issue.

ajCameron: Added as issue https://github.com/mewler/mewbot/issues/173

javajawa:
I've updated this based on the work in #197 (so it is now a 'normal' toolchain) but haven't actually reviewed the generator code properly. However, I think anything I have left is going to be pretty minor -- the only thing that stood out during rebasing was the logic to generate an index turns up in about 3 places.

exachixkitsune (dismissed):
Seems ok, though probably wants @javajawa's comment about duplication addressed?

javajawa (dismissed):
Discusses in person: the way indexes are generated isn't elegant but is not a blocker. We can raise an issue to explore solutions like having an index Object that is built up by various methods are rendered out in a consistent method.

javajawa (approved):

PR #172, opened 2023-03-27
Upgrading CI linting tools

ajCameron:
- added capture and printing of stdout and stderr during ci run
 - adding timeout functionality to prevent process hang

### DIFF
--- a/src/mewbot/tools/__init__.py
+++ b/src/mewbot/tools/__init__.py
@@ -9,12 +9,15 @@
 from __future__ import annotations
 
 import abc
+import asyncio
 import dataclasses
 import os
 import subprocess
 import sys
-
 from collections.abc import Iterable
+from io import BytesIO
+
+from typing import IO, BinaryIO
 
 from .path import gather_paths
 
@@ -67,6 +70,8 @@ class ToolChain(abc.ABC):
     is_ci: bool
     success: bool
 
+    timeout: int = 300
+
     def __init__(self, *folders: str, in_ci: bool) -> None:
         """
         Sets up a tool chain with the given settings.
@@ -78,8 +83,15 @@ class ToolChain(abc.ABC):
         self.is_ci = in_ci
         self.success = True
 
+        self.loop = asyncio.get_event_loop()
+
     def __call__(self) -> None:
         """Runs the tool chain, including exiting the script with an appropriate status code."""
+        # Windows hack to allow colour printing in the terminal
+        # See https://bugs.python.org/issue30075 and our windows-dev-notes.
+        if os.name == "nt":
+            os.system("")
+
         # Ensure the reporting location exists.
         if not os.path.exists("./reports"):
             os.mkdir("./reports")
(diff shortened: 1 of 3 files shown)


src/mewbot/tools/__init__.py
```diff
@@ -147,10 +149,68 @@ def _run_utility(
...
+                        else:
+                            # POSIX _communicate already populated the output so
+                            # far into the TimeoutExpired exception.
+                            process.wait()
+                        raise
+                    except Exception as exc:
+                        # Including KeyboardInterrupt, communicate handled that.
+                        process.kill()
+                        # We don't call process.wait() as .__exit__ does that for us.
+                        raise NotImplementedError(
```
javajawa:
```suggestion
                        raise TimeoutError(
```

javajawa:
Been talking a look at this, I'm not sure what the overall change works out to?

It chnages the behaviour if `self.is_ci`, but then the code appears to be identical to the `subprocess.run` when `capture_output` is true...which it already will be? The code difference in a while loop, but the what `process.communicate()` works is that's is a blocking call that waits for the entire output.

I have had some success with code like

```python

    async def _run_utility(self, name: str, arg_list: list[str]) -> CompletedProcess[bytes]:
        """
        Helper function to run an external program as a check.

        The output of the command is made available in three different ways:
          - Output is written to reports/{tool name}.txt
          - Output and Error are returned to the caller
          - Error and Output are copied to the terminal.

        When in CI mode, we add a group header to collapse the output from each
        tool for ease of reading.

        :param name: The user-friendly name of the tools
        :param arg_list: The command line to use. The first value should be the executable path.
        """

        # Print output header
        print(f"::group::{name}" if self.is_ci else f"\n{name}\n{'=' * len(name)}")

        process = await asyncio.create_subprocess_exec(
            *arg_list, stdin=DEVNULL, stdout=PIPE, stderr=PIPE
        )

        # MyPy validation trick -- ensure the pipes are defined (they will be).
        if not process.stdout or not process.stderr:
            raise ValueError(f"pipes for process {name} not created")

        # Set up the mirroring readers on the two output pipes.
        task_out = self.loop.create_task(self.read_pipe(process.stdout, sys.stdout.buffer))
        task_err = self.loop.create_task(self.read_pipe(process.stderr, sys.stderr.buffer))

        # This is trimmed down version of subprocess.run().
        try:
            await asyncio.wait_for(process.wait(), timeout=self.timeout)
        except TimeoutError:
            process.kill()
            # TODO: run uses communicate() on windows. May be needed.
            #   However, as we are running the pipes manually, it may not be.
            await process.wait()
        except:  # Re-raise all non-timeout exceptions.
            process.kill()
            await process.wait()
            raise
        finally:  # Ensure the other co-routines complete.
            stdout_buffer = await task_out
            stderr_buffer = await task_err

        retcode = process.returncode
        retcode = retcode if retcode is not None else 1

        run = CompletedProcess(arg_list, retcode, stdout_buffer.read(), stderr_buffer.read())

        with open(f"reports/{name}.txt", "wb") as log_file:
            log_file.write(run.stdout)

        if self.is_ci:
            print("::endgroup::")

        return run

    @staticmethod
    async def read_pipe(pipe: asyncio.StreamReader, mirror: BinaryIO) -> IO[bytes]:
        """
        Read a pipe from a subprocess into a buffer whilst mirroring it to another pipe.
        """

        buffer = BytesIO()

        while not pipe.at_eof():
            block = await pipe.readline()
            mirror.write(block)
            mirror.flush()
            buffer.write(block)

        buffer.seek(0)
        return buffer
```

but it needs testing with configurations I don't have access to (namely Windows).

The test I'm using locally looks as follows

```python
from . import ToolChain

class TestToolchain(ToolChain):
    def run(self):
        self.timeout = 8
        result = self.run_tool("Count", "/bin/bash", "-c", "X=0; while sleep 1; do X=$(( X + 1 )); echo $X; done")

        print(result)
        assert result.returncode != 0
        assert len(result.stdout.split(b"\n")) == 7  # First second has no print.
        yield None


if __name__ == "__main__":
    TestToolchain(in_ci=False)()
```

ajCameron:
Ah - I see the confusion.

I mentioned to @javajawa in conversation that this solution was capturing the output live. I was misremembering. This turned out to be hard, so I had to adopt another solution.

I compromised on "capture the output and kill the process after a while". Which seemed to have most of the benefits of live capture (getting the output from the failed process) and the added benefit of imposing a timeout on each utility run, so we wouldn't waste resources e.g. if a test in the suite falls into an infinite while loop.

@javajawa's solution looks more elegant, and preserves the timeout, so I will test on windows and adopt it.

src/mewbot/tools/lint.py
```diff
@@ -190,13 +190,19 @@ def lint_black_errors(
...
     errors = result.stderr.decode("utf-8").split("\n")
+
     for error in errors:
         error = error.strip()
 
         if not error:
             continue
 
-        level, header, message, line, char, info = error.split(":", 5)
+        # Black is getting confused as to what an error is on windows
```
javajawa: I think I solved this with `if not error or ":" not in error:`.

src/mewbot/tools/__init__.py
```diff
@@ -123,49 +134,128 @@ def run_tool(self, name: str, *args: str) -> subprocess.CompletedProcess[bytes]:
...
+        run_result = self.loop.run_until_complete(self._async_run_utility(name, arg_list))
+
+        self.success = self.success and (run_result.returncode == 0)
+        assert isinstance(run_result, subprocess.CompletedProcess)
+        return run_result
+
+    async def async_run_tool(
+        self, name: str, *args: str
+    ) -> subprocess.CompletedProcess[bytes]:
+        """
```
javajawa: Don't duplicate the logic here, just have `run_tool` be
```
return self.loop.run_until_complete(self.async_run_tool(name, *args)
```

Or just make `run()` and `run_tool()` always async, move the loop logic up into `__call__`, and that means that tools can easily set up their own concurrency.

src/mewbot/tools/__init__.py
```diff
@@ -123,49 +134,128 @@ def run_tool(self, name: str, *args: str) -> subprocess.CompletedProcess[bytes]:
...
+
+        run_result = await self._async_run_utility(name, arg_list)
+        assert isinstance(run_result, subprocess.CompletedProcess)
 
         self.success = self.success and (run_result.returncode == 0)
 
         return run_result
 
-    def _run_utility(
+    async def _async_run_utility(
```
javajawa: No real need to change the name here, there's no sync equivalent remaining.

src/mewbot/tools/__init__.py
```diff
@@ -78,6 +83,12 @@ def __init__(self, *folders: str, in_ci: bool) -> None:
         self.is_ci = in_ci
         self.success = True
 
+        self.loop = asyncio.get_event_loop()
+
+        # Windows hack to allow colour printing in the terminal - for mypy at least
+        if os.name == "nt":
+            os.system("")
```
javajawa: This should probably be in `__call__`, so the world isn't changed until the tool chain is actually run.

javajawa (approved):

PR #176, opened 2023-04-09
Combine CI lint and test matrix

javajawa:
This resuces the amount of code in the CI definition,  and highlights problems caused by different linting behaviours on different versions of python

### DIFF
--- a/src/mewbot/tools/__init__.py
+++ b/src/mewbot/tools/__init__.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import abc
 import asyncio
 import dataclasses
+import json
 import os
 import subprocess
 import sys
@@ -32,6 +33,7 @@ class Annotation:
     file: str
     line: int
     col: int
+    tool: str
     title: str
     message: str
 
@@ -41,7 +43,19 @@ class Annotation:
         mess = self.message.replace("\n", "%0A")
         return (
             f"::{self.level} file={self.file},line={self.line},"
-            f"col={self.col},title={self.title}::{mess}"
+            f"col={self.col},title={self.title.strip()}::{mess.strip()}"
+        )
+
+    def json(self) -> dict[str, str | int]:
+        """Output this object as a JSON-encodeable dictionary."""
+
+        return dataclasses.asdict(self)
+
+    def __hash__(self) -> int:
+        """Unique hash of this annotation."""
+
+        return hash(
+            (self.level, self.file, self.line, self.col, self.tool, self.title, self.message)
         )
 
     def __lt__(self, other: Annotation) -> bool:
@@ -98,8 +112,10 @@ class ToolChain(abc.ABC):
 
         issues = list(self.run())
 
-        if self.is_ci:
-            self.github_list(issues)
+        with open(
+            f"./reports/annotations-{self.__class__.__name__}.json", "w", encoding="utf-8"
+        ) as output:
+            json.dump([issue.json() for issue in issues], output, indent=2)
 
         sys.exit(not self.success or len(issues) > 0)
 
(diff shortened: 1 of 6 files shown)


javajawa:
I've remembered why I don't normally do this: there's no de-duplication on annotations, so a lot of things will get annotated 9 different times. We could build infrastructure around that, or continue only linting one version.

(Infra would probably be writing the annotations out to a file in each linting stage, then in a later stage with `if: always()` grouping by file+line and then posting one annotation with each unique message. This would also improve the behaviour of having multiple annotations on one line)

javajawa:
I'm going to sort out #172 before addressing https://github.com/mewbotorg/mewbot/pull/176#issuecomment-1502863235, but I think the idea will indeed to be shift it around as artifacts

javajawa:
There's some tidy up still needed, but as this changes the way in which annotations are applied it should be reviewed whilst there are still annotations.

Examples can be seen such as https://github.com/mewbotorg/mewbot/pull/176/commits/f0b429b744e31ebc66ce7b71de2e996270683cec#diff-af4cde4b1869c031eae821356802a3e2e0d62aba4cf0b5694722434ae923bdccR48

ajCameron:
This looks really good!

Annotations do, indeed, seem to be applying correctly. I had not considered the multi-annotation issue, but indeed that's a serious problem and the proposed fix is a good one.

Couple of suggestions - more nice to haves than anything else
 - it'd be _really_ cool if annotation was provided with some details as to which linter produced it and the python version it was running under at the time (e.g. - the annotation itself prepended with "black-3.9/3.10" e.t.c.)
 - some better names for the tests and linting jobs would be good - but, again, a nice to have

src/mewbot/tools/__init__.py
```diff
@@ -98,8 +105,10 @@ def __call__(self) -> None:
 
         issues = list(self.run())
 
-        if self.is_ci:
-            self.github_list(issues)
+        with open(
```
exachixkitsune, line 115: Query: Is this something we want to do on every tool run, or only github? (as previously it was only operating when in github)

src/mewbot/tools/annotate.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: 2021 - 2023 Mewbot Developers <mewbot@quicksilver.london>
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Output Annotations from previous CI stages into CI
```
exachixkitsune:
```suggestion
Output Annotations from previous CI stages into CI.
```

src/mewbot/tools/annotate.py
```diff
@@ -0,0 +1,99 @@
...
+from collections.abc import Iterable
+
+import os
+
+from . import Annotation, ToolChain
+
+
+class Annotate(ToolChain):
+    """
+    Output Annotations from previous CI stages into CI
```
exachixkitsune:
```suggestion
    Output Annotations from previous CI stages into CI.
```

src/mewbot/tools/annotate.py
```diff
@@ -0,0 +1,99 @@
...
+
+            title = f"{len(issues)} Issues on this line"
+            message = "\n\n".join(self.format_sub_issue(issue) for issue in issues)
+
+            yield Annotation(level, file, line, 0, title, message)
+
+    @staticmethod
+    def format_sub_issue(issue: Annotation) -> str:
+        """
+        Converts an existing annotation
```
exachixkitsune:
```suggestion
        Converts an existing annotation.
```

exachixkitsune (dismissed):
Looks good! I would prehaps suggest some additional commentry; it took a bit to work out what group_issues was doing for example - but this is the area of code people may not look at so much so ¯\_(ツ)_/¯

src/mewbot/tools/__init__.py
javajawa: It was previously only in CI to stop the annotations being output locally, as they're not exactly readable. But as they now go to file, the need for different behaviour is diminished.

After these changes, the differences enabling the CI flag makes are:
 - `black` outputs a diff rather than auto fixing.
 - `mypy` has colour and pretty output disabled
 - `pytest` does not generate HTML coverage

javajawa:
Because this one is weird with the CI always looking broken: this *is* ready for review and merging.

Once approved, the `Lint Python Code` CI step will have to be removed from the requires step list (as it no longer exists). The linters passing is now aggregated into the `Auto-Review / Tests Passed` step, which is also marked as required.

PR #190, opened 2023-04-25
Hack to improve python version compatibility

ajCameron:
- added an if to the path script to try and use the right one from "python" and "python3"
 - for improved windows compatibility

Changes to path from https://github.com/mewbotorg/mewbot/pull/49.

That PR is now fully deprecated and will be removed.

### DIFF
--- a/tools/path
+++ b/tools/path
@@ -4,7 +4,16 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-PYTHON=${PYTHON:-python3}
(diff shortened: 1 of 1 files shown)


javajawa:
It might be overkill, but here's a posix shell implementation that checks the version of `python` https://github.com/mewbotorg/mewbot/compare/python_path_auto_detect?expand=1

ajCameron:
@javajawa - looks good!

No kill like overkill. Created a pull request for it and closing this one.

PR #195, opened 2023-05-16
Mewbot discord dice roller example plugin

### LINKED ISSUE #182: Create plugin for `discord_dice_roller`
#120 added code for the discord_dice_roller module.

This request was refactored into a series of smaller requests, none of which included that code. It should be rebased and resubmitted for approval.

ajCameron:
- aims to resolve https://github.com/mewbotorg/mewbot/issues/182
 - minor correction to the windows dev install docs
 - added example of direct import - access through registry needs more work than should be accommodated in this pull request
 - added mewbot-discord_dice_roller plugin with example
 - upgraded examples to allow running of examples stored in builtin plugins via path manipulation
 - added an "examples" sh script - through which you can run the examples
 - added notes to API_DISTRIBUTIONS constant in registry
 - fixed weird issue in bot which was preventing run - on windows - with some python versions

Still todo
 - sort out packaging for modules
 - write an interface for registry to allow direct listing and loading of modules from there
however, including these would grow this pull request beyond reasonable scope

### DIFF
--- a/dev-docs/plugin-dev-notes/README.md
+++ b/dev-docs/plugin-dev-notes/README.md
@@ -16,7 +16,7 @@ It's designed to take yaml files, load them, and turn them into bots which can i
 While it does have some basic capabilities, the majority of its capabilities come from plugins
 (either plugins already bundled with the program or plugins written for it.)
 
-Thus, if you want to extend the capabilities of the _framework itself_ you probably want to write a **mewbot framework plugin**.
+If you want to extend the capabilities of the _framework itself_ you want to write a **mewbot framework plugin**.
 
 On a technical level this will take the form of a namespace plugin which will use the `mewbot_framework_plugin_template`.
 
@@ -25,9 +25,21 @@ I will write a plugin which will present a new IOConfig in `mewbot.io`
 
 For further information on how to write namespace plugins, please see the python core documentation - [python namespace plugin docs](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/).
 
+Your plugin will be used by installing your module and then calling the IOConfig using something like
+
+```yaml
+kind: IOConfig
+implementation: mewbot.io.mastadon.MastadonIO
+uuid: aaaaaaaa-aaaa-4aaa-0000-aaaaaaaaaa00
+properties:
+  token: "[your mastadon token goes here]"
+
+---
+```
+
 #### 2) Creating tools for bots
 
-If, on the other hand, you want to write some tools which bots _themselves_ will use, then you probably want to write a **mewbot bot plugin**.
+If, on the other hand, you want to write some tools which bots _themselves_ will use, then you may want to write a **mewbot bot plugin**.
 
 This would be a plugin for mewbot which does not add to the core module namespace.
 It might include detailed Triggers, Conditions and Actions that you will use to build your bot via yaml.
@@ -45,30 +57,30 @@ As a rule of thumb
 
 E.g.
 
-#### Example 1
+#### Example 1 - "I want to allow mewbot to talk to mastadon."
 
-"I want to allow mewbot to talk to mastadon"
-
-Probably write a **framework** plugin.
+Write a **framework** plugin.
 
 In fact, please do!
 That would really be very helpful.
 
 Your generic IOConfig will take the form of a namespace plugin.
-It will be used by importing it using something like
+It will be used by importing it from the wider mewbot namespace via something like this
 
 ```python
 from mewbot.io.mastadon import MastadonIOConfig
 ```
 
-#### Example 2
+It will, to all intents and purposes, serve as part of mewbot itself.
 
-"I've written a discord dice roller! I'd like to distribute it."
+#### Example 2 - "I've written a discord dice roller! I'd like to distribute it."
 
-Probably write a **bot** plugin.
+Write a **bot** plugin.
 
-Structure it how you like.
-Use the components by having lines such as 
+Structure it how you like - it matters a lot less for this sort of plugin than a namespace plugin.
+(Though there is an example in plugins under `mewbot-discord_dice_roller` which should serve to get you started).
+
+Use the components in your bot by having lines such as
 
 ```yaml
 kind: Behaviour
@@ -94,13 +106,27 @@ actions:
 
 ```
 
+in your yaml files.
+
 (descriptive names are preferred for this sort of plugin as the yaml is intended to be highly human readable).
 
-#### Example 3
+#### Example 3 - "I've written some awesome conditions. I want to distribute them."
 
-"I've written some awesome conditions. I want to distribute them."
-
-You probably want a **bot** plugin.
+You may want a **bot** plugin.
 But, if they are _extremely_ generic, they might be candidates for inclusion in `mewbot.io.common`.
 This will, however, be fairly rare.
 To keep the core framework as light weight as possible.
+
+#### Example 4 - "I want to help out by writing a plugin. What sort of plugin should I write?"
+
+As a rule of thumb
+ - **framework** plugins are harder to write, but more helpful
+ - **bot** plugins are easier to write, but less helpful
+
+Framework plugins allow mewbot bots to talk to more services.
+Which, for every service, vastly increases the number of useful bots than can be written.
+
+Bot components allow the bots to do more things.
+They provide capabilities, and allow each bot written for mewbot to do more.
+
+Both are very helpful!
--- a/dev-docs/plugin-dev-notes/README.md
+++ b/dev-docs/plugin-dev-notes/README.md
@@ -1,0 +1,106 @@
+
+# Plugin dev notes
+
+Notes on creating plugins and extensions for mewbot.
+
+## Plugin writing intro
+
+### Why you might want to write a plugin?
+
+There are two reasons - and two types of plugin you might want to write.
+
+#### 1) Extending mewbot itself
+
+Base mewbot is more a framework than a complete program.
+It's designed to take yaml files, load them, and turn them into bots which can interact with the world.
+While it does have some basic capabilities, the majority of its capabilities come from plugins
+(either plugins already bundled with the program or plugins written for it.)
+
+Thus, if you want to extend the capabilities of the _framework itself_ you probably want to write a **mewbot framework plugin**.
+
+On a technical level this will take the form of a namespace plugin which will use the `mewbot_framework_plugin_template`.
+
+E.g. I would like mewbot to be able to talk to "Insert communications protocol here".
+I will write a plugin which will present a new IOConfig in `mewbot.io`
+
+For further information on how to write namespace plugins, please see the python core documentation - [python namespace plugin docs](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/).
+
+#### 2) Creating tools for bots
+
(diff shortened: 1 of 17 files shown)


dev-docs/plugin-dev-notes/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+In fact, please do!
+That would really be very helpful.
+
+Your generic IOConfig will take the form of a namespace plugin.
+It will be used by importing it using something like
+
+```python
+from mewbot.io.mastadon import MastadonIOConfig
+```
```
exachixkitsune:
```suggestion
"I want to allow mewbot to talk to mastodon"

Probably write a **framework** plugin.

In fact, please do!
That would really be very helpful.

Your generic IOConfig will take the form of a namespace plugin.
It will be used by importing it using something like

```python
from mewbot.io.mastodon import MastodonIOConfig
```
```

File: dev-docs/plugin-dev-notes/README.md [hunk 2]
```diff
@@ -0,0 +1,106 @@
...
+
+#### Example 2
+
+"I've written a discord dice roller! I'd like to distribute it."
+
+Probably write a **bot** plugin.
```
exachixkitsune:
```suggestion
Write a **bot** plugin.
```

It is my opinion these should be definitives rather than have wiggle room; but that's potentially on me.
Would there be a reason we make a discord dice roller as part of the framework vs a plugin?

plugins/mewbot-discord_dice_roller/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Ho do I use these components?
+
+There are two ways to do this
+
+1) Import the components directly from the `mewbot_discord_dice_roller` module
+2) Access them through the registry
+
+Which of these you use is up to you!
+
+Included in this module is two examples - one for each method.
```
exachixkitsune, line 42: I would also include what needs to go into the `token: "[token goes here]"` part of the `direct_import_dice_roller_bot.yaml` file.

plugins/mewbot-discord_dice_roller/setup.py
```diff
@@ -0,0 +1,75 @@
...
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Note - this is NOT intended to be used for local installation or as part of development.
+
+This is intended to make this module available via pypi
+It _should_ also work for local installation, but this functionality is not tested.
+"""
+
+raise NotImplementedError("Still under development.")
```
exachixkitsune, line 14: Do we want this implemented for the purpose of this demo? or is it dependant on other things?

plugins/mewbot-discord_dice_roller/src/mewbot_discord_dice_roller/behaviors.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+
+"""
+This Behavior brings all the componets for the Discord Dice roller together in one place.
```
exachixkitsune:
```suggestion
This Behavior brings all the components for the Discord Dice roller together in one place.
```

plugins/mewbot-discord_dice_roller/src/mewbot_discord_dice_roller/triggers.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def matches(self, event: InputEvent) -> bool:
+        """
+        Matches all DiscordMessageCreationEvent with the starting keyword !roll.
+
+        :param event:
+        :return:
+        """
+
+        if not isinstance(event, DiscordMessageCreationEvent):
```
exachixkitsune, line 40: Idle thought for future code; is there a way to make sure this happens automatically/is not forgettable by a user?
Is there a reason not to do that?

exachixkitsune (dismissed):
Looking good, comments are mostly ignorable, spelling, or idle future thoughts.

dev-docs/plugin-dev-notes/README.md
ajCameron: I've gone through and re-written this section to be more definite "probably write" has become "write" - in line with the suggestion below.

plugins/mewbot-discord_dice_roller/setup.py
javajawa: It's...a little complicated. We're already doing some pretty nonstandard things with `setup.py` in other places with the multi-module build system. Probably could make it work right now, but I'm happy to ignore it.

I'm actually going to suggest (when we preoperly get there) using `pyproject.toml` rather than `setup.py` for the plugins -- that uses more sensible defaults and means there isn't a random python file for installation.

plugins/mewbot-discord_dice_roller/src/mewbot_discord_dice_roller/triggers.py
javajawa: The [Liskov substitution principle](https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides) boils down to the argument to the function being `InputEvent` (or `OutputEvent` for `Output`s), so the check is going to be needed for typing purposes. It's also just a good defensive measure against poorly implemented bot runners.

It _is_ a little cumbersome, and we could abstract it to a decorator...
A fully generic version might look like

```python
from typing import Callable, ParamSpec, TypeVar

import mewbot.api.v1
from mewbot.core import InputEvent, OutputEvent
from mewbot.io.common import EventWithReplyMixIn


Args = ParamSpec("Args")
Bound = TypeVar("Bound", bound=InputEvent)
Return = TypeVar("Return")


def check_types(*types: type[Bound], result: Return = ...) -> Callable[[Callable[Args, Return]], Callable[Args, Return]]:
    def check_wrapper(wrapped: Callable[Args, Return]) -> Callable[Args, Return]:
        def inner(*args: Args.args, **kwargs: Args.kwargs) -> Return:
            print(args)
            if not isinstance(args[1], types):
                if result is ...:
                    raise TypeError(f"First argument to {wrapped} must be one of {types}")
                return result

            return wrapped(*args, **kwargs)

        return inner
    return check_wrapper


class A(mewbot.api.v1.Trigger):
    @staticmethod
    def consumes_inputs() -> set[type[InputEvent]]:
        return {EventWithReplyMixIn}

    @check_types(EventWithReplyMixIn, result=False)
    def matches(self, event: InputEvent) -> bool:
        return True
```

I suspect one which is configured specifically for the different functions would, however, be more useful (and not make mypy scream)

... ok, I took some time to play and it is 100% possible:

```python
from __future__ import annotations

from typing import Callable, TypeVar

from mewbot.api.v1 import Trigger, Condition, InputEvent
from mewbot.io.discord import DiscordInputEvent, DiscordUserJoinInputEvent

Component = TypeVar("Component", bound=Trigger | Condition)
Event = TypeVar("Event", bound=InputEvent)


def check_types(wrapped: Callable[[Component, Event], bool]) -> Callable[[Component, InputEvent], bool]:
    def inner(self: Component, event: InputEvent) -> bool:
        if not isinstance(event, tuple(self.consumes_inputs())):
            return False

        return wrapped(self, event)  # type: ignore

    return inner


class A(Trigger):
    @staticmethod
    def consumes_inputs() -> set[type[InputEvent]]:
        return {DiscordInputEvent}

    @check_types
    def matches(self, event: DiscordInputEvent) -> bool:
        print(event)
        return True


print("invoking A()")
print(A().matches(InputEvent()))
print(A().matches(DiscordInputEvent()))
print(A().matches(DiscordUserJoinInputEvent("nya")))
```

That passes typing checks; there is one ignore there because I can't _prove_ that `Bound` is implied by the `isintance` call above.

javajawa: ...if we want to get really, really clever, we could actually shift the concern of `consume_inputs` to the type signature of `matches`...

I think this is where it veers into bad idea territory. And I can't easily make it work, but I have made a slightly better version of the above (which doesn't need any mypy suppressions, but *does* confuse PyCharm a bit):
```python
from __future__ import annotations

import functools

from types import UnionType
from typing import Callable, TypeVar, get_type_hints

from mewbot.api.v1 import Trigger, Condition, InputEvent

Component = TypeVar("Component", bound=Trigger | Condition)
Event = TypeVar("Event", bound=InputEvent)


def check_types(
    wrapped: Callable[[Component, Event], bool]
) -> Callable[[Component, InputEvent], bool]:
    types = get_type_hints(wrapped)
    if "event" not in types:
        raise TypeError("Received function without 'event' parameter")

    event_types: type[Event] = types["event"]
    if not isinstance(event_types, (type, UnionType)):
        raise TypeError("Received function with invalid type for 'event'")

    @functools.wraps(wrapped)
    def inner(self: Component, event: InputEvent) -> bool:
        # noinspection PyTypeHints
        if not isinstance(event, event_types):
            return False

        return wrapped(self, event)

    return inner


# Demo Code

from mewbot.io.discord import DiscordInputEvent, DiscordUserJoinInputEvent
from mewbot.io.http import IncomingWebhookEvent


class A(Trigger):
    @staticmethod
    def consumes_inputs() -> set[type[InputEvent]]:
        return {DiscordInputEvent, IncomingWebhookEvent}

    @check_types
    def matches(self, event: DiscordInputEvent | IncomingWebhookEvent) -> bool:
        """Hello I am documentation"""
        return True


print(A.matches)  # <function A.matches at 0x7f6e703f4a40>
print(A.matches.__doc__)  # <function A.matches at 0x7f6e703f4a40>
print(A().matches(InputEvent()))  # False
print(A().matches(DiscordInputEvent()))  # True
print(A().matches(DiscordUserJoinInputEvent("nya")))  # True
```

And I think we could store the `UnionType` somewhere and use it for `consume_inputs` if we really wanted to. But I also feel that might be Too Much Magic:tm:

dev-docs/plugin-dev-notes/README.md
ajCameron: Agreed - implemented throughout the docs.

plugins/mewbot-discord_dice_roller/README.md
ajCameron: Included and referenced in the places which seem to make sense.

plugins/mewbot-discord_dice_roller/setup.py
ajCameron: I'm gonna have a think and a poke some, and see what I come up with.

plugins/mewbot-discord_dice_roller/src/mewbot_discord_dice_roller/behaviors.py
ajCameron: Corrected

plugins/mewbot-discord_dice_roller/src/mewbot_discord_dice_roller/triggers.py
ajCameron: I really like this. It's very elegant and could be more magical.

This feels like it should be it's own pull request. I'll do one tomorrow to get it into the code base and get it integrated.

exachixkitsune (approved):
Looks good! Extent conversations - probably need you to decide if they are solved, or deferred to other PRs.

plugins/mewbot-discord_dice_roller/setup.py
ajCameron: I have something which works, but feel it's outside the scope of this pull request.

PR #196, opened 2023-05-16
Adding preflight skeleton

ajCameron:
Contains the programmatic parts of https://github.com/mewbotorg/mewbot/pull/175 - which is deprecated by this pull request.

 - preflight is intended to be a single script you run before submitting a pull request
 - it runs reuse, linters, then the tests in order and reports on the result
 - reuse has been added to tools as a script - and to mewbot.tools as a runnable module
 - terminal module has been added to tools with some helpful delineation functions


src/mewbot/tools/lint.py
```diff
@@ -38,14 +39,24 @@ class LintToolchain(ToolChain):
...
+            yield from self.lint_flake8()
+        with CommandDeliminator("Starting mypy tool run"):
+            yield from self.lint_mypy()
+        with CommandDeliminator("Starting pylint tool run"):
+            yield from self.lint_pylint()
+        with CommandDeliminator("Starting pydocstyle tool run"):
+            yield from self.lint_pydocstyle()
+
+        if self.is_ci and self.fail_ci_run:
+            print(f"Some of the tools reports having silently failed!\n{self.run_success}")
```
exachixkitsune: What's stopping this from appearing if something hasn't silently failed? I cannot see where that would occur.

exachixkitsune (dismissed):
Only one minor comment; looks good to me!

javajawa (changes requested):
Loving the pre-flight part, I'm going to have a little play around combining the `CommandDeliminator` with the existing login in the `_run_utility` function -- the pipelines are looking a little messier: https://github.com/mewbotorg/mewbot/actions/runs/4993044197/jobs/8941766975

javajawa:
I have a series of proposals that are a little too complicated for pull request comments, so I have opened a dependant PR #197 with my thinking.

The first bits are minor fixes/corrections:
 - Consistency Fix: is_ci -> in_ci (not really related to this PR, but it made things easier)
 -  Mark `tools/preflight` as executable (a little `chmod` action)
 - Push the `CommandDeliminator` call down into `ToolChain._run_utility`
   - I toyed around here with the idea of it prefix each output line with the command's name
 - Put a loop call when calling the test runs so preflight actually runs them.

Then come the big changes:
 - Refactor `ReuseRun` to `ReuseToolchain` and later add `reuse lint` as a step.
 - Refactor `PreFlight` to a ToolChain, to take advantage of code reuse.
 - Change the big Report format to looks like 
2023-05-20_300x226
 - Change the command delimitor to look like 
2023-05-20_666x158
 - In CI, it looks like 2023-05-20_693x467

@ajCameron if you're happy with all that, you can fast-forward this branch to that one, and GitHub will mark 197 as merged without any need for a merge commit :)

javajawa (approved):

src/mewbot/tools/lint.py
javajawa: This has been removed :)

PR #202, opened 2023-05-22
Improve handling in tools/path

javajawa:
The logic to detect the python binary path has been enhanced by seperating user-communication from the execution flow, and making use of shell if-statement features that were recommended by shell-check.

This change was inspired during a rebase of #169

### DIFF
--- a/tools/path
+++ b/tools/path
@@ -4,48 +4,51 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-
 auto_detect_python()
 {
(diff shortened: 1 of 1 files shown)


javajawa:
I've tested heavily on Linux and it's all working well. The features I'm using are present in `dash`, which is the closest thing I have to a pure posix shell. They're also present in `zsh` and `bash`, which should cover most configurations.

ajCameron (approved):
Looks good and working well on windows!

PR #203, opened 2023-05-23
Adding the type guard decorator

ajCameron:
- in api - seemed the most appropriate place for it
 - (if nothing else, because it needed less cross module imports)
 - added extensive docs including example code
 - selected the option which didn't confuse pycharm - as that seems to be harder to fix
 - included a test suite hitting 100% of the module
 - decorator not currently applied to the examples - would seem beyond the scope of a single pull request
 - Not over sold on the name - but felt descriptive was better

### DIFF
--- a/tests/api/type_guard/test_type_guard_invalid_config.py
+++ b/tests/api/type_guard/test_type_guard_invalid_config.py
@@ -1,0 +1,164 @@
+"""
+Test attempting to use the type guard for Triggers and Conditions in various invalid configurations.
+"""
+
+from __future__ import annotations
+
+import types
+import typing
+from typing import Optional, Union
+
+import pytest
+
+from mewbot.api.v1 import pre_filter_non_matching_events
+from mewbot.core import InputEvent
+from mewbot.io.http import IncomingWebhookEvent
+from mewbot.io.socket import SocketInputEvent
+
+union_type: Optional[type] = getattr(types, "UnionType", None)
+
+TYPE_ERROR_MESSAGE = r".*.matches: Can not add filter for non-event type\(s\): .*"
+
+
+class TestCheckInputEventAgainstSigTypesAnnotations:
+    """
+    Check the type guard annotation check_input_event_against_sig_types preforms as expected.
+    """
+
+    def test_union_syntax_on_different_versions(self) -> None:
+        """
+        Probe annoyance where the union syntax seems to yield different results on 3.9 and 3.10.
+
+        Preserved here as a canary for when a future version of python changes this behavior.
+        Which will hopefully be soon.
+        Either I'm missing something, or it's stupid.
+        :return:
+        """
+        assert union_type is not None, "union_type was None - change to python syntax?"
+
+        # This is not valid syntax in python 3.9 - so need to ignore it for mypy
+        test_syntax_1 = SocketInputEvent | IncomingWebhookEvent
+        # pylint: disable=isinstance-second-argument-not-valid-type
+        assert isinstance(test_syntax_1, union_type)
+        # pylint: disable=protected-access
+        assert not isinstance(test_syntax_1, typing._UnionGenericAlias)  # type: ignore
+
+        test_syntax_2 = Union[SocketInputEvent, IncomingWebhookEvent]
+        # You would really hope this would work ... but, alas, it does not
+        # pylint: disable=isinstance-second-argument-not-valid-type
+        assert not isinstance(test_syntax_2, union_type)
+        # pylint: disable=protected-access
+        assert isinstance(test_syntax_2, typing._UnionGenericAlias)  # type: ignore
+
+    def test_malformed_trigger_no_event_annotations(self) -> None:
+        """
+        Tests applying the decorator to a function without "event" as a possibility.
+
+        :return:
+        """
+        with pytest.raises(TypeError, match="Received function without 'event' parameter"):
+
+            @pre_filter_non_matching_events  # type: ignore
+            def consumes_inputs() -> set[type[InputEvent]]:
+                """
+                Consumes two radically different events.
+
+                :return:
+                """
+                return {SocketInputEvent, IncomingWebhookEvent}
+
+    def test_malformed_trigger_bad_event_type_none_annotations(self) -> None:
+        """
+        Tests applying the decorator to a function without "event" as a possibility.
+
+        :return:
+        """
+
+        with pytest.raises(TypeError, match=TYPE_ERROR_MESSAGE):
+
+            @pre_filter_non_matching_events  # type: ignore
+            def matches(event: None) -> bool:
+                """
+                If the event is of the right type, matches.
+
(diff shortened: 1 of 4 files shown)


src/mewbot/api/type_guard.py
```diff
@@ -0,0 +1,97 @@
...
+
+"""
+
+import functools
+
+from typing import Callable, TypeVar, get_type_hints, Union
+
+from mewbot.api.v1 import Trigger, Condition, InputEvent
+
+Component = TypeVar("Component", bound=Union[Trigger, Condition])
```
javajawa: If we're doing this 'properly', we should either move this into `mewbot.api.v1` (possible as a re-exported import from a sub-module), or change these to use the `*Interface` protocols from `mewbot.core`.

src/mewbot/api/type_guard.py
```diff
@@ -0,0 +1,97 @@
...
+        - If it is not, False is returned
+    :param wrapped:
+    :return:
+    """
+    types = get_type_hints(wrapped)
+    if "event" not in types:
+        raise TypeError("Received function without 'event' parameter")
+
+    event_types: type[Event] = types["event"]
+    if not isinstance(event_types, (type, type(Union))):
```
javajawa: I also don't know if this is a valid transform from `UnionType`, but I think it might be? If the tests pass after my next comment, it should be fine.

tests/api/test_type_guard.py
```diff
@@ -0,0 +1,210 @@
...
+    def consumes_inputs() -> set[type[InputEvent]]:
+        """
+        Consumes two radically different events.
+
+        :return:
+        """
+        return {SocketInputEvent, IncomingWebhookEvent}
+
+    @check_input_event_against_sig_types
+    def matches(self, event: Union[SocketInputEvent, IncomingWebhookEvent]) -> bool:
```
javajawa: We should have at least one test using `X | Y` syntax because it is slightly different. `from __future__ import annotations` should allow it in python 3.9, but I might recommend having a separate test file for it -- that we can have better confidence that both forms work as expected.

src/mewbot/api/type_guard.py
```diff
@@ -0,0 +1,97 @@
...
+        raise TypeError("Received function with invalid type for 'event'")
+
+    @functools.wraps(wrapped)
+    def inner(self: Component, event: InputEvent) -> bool:
```
javajawa: We could go more verbose here (`wrapped` -> `user_supplied_function`, `inner` -> `api_compatible_wrapper`... I don't like _any_ of these names.)

src/mewbot/api/type_guard.py
```diff
@@ -0,0 +1,97 @@
...
+
+from typing import Callable, TypeVar, get_type_hints, Union
+
+from mewbot.api.v1 import Trigger, Condition, InputEvent
+
+Component = TypeVar("Component", bound=Union[Trigger, Condition])
+Event = TypeVar("Event", bound=InputEvent)
+
+
+def check_input_event_against_sig_types(
```
javajawa: I'm ok with this name, but if you're still unsure (from the commit messages) another naming approach might be along the lines of `pre_filter_non_matching_events`?

src/mewbot/api/type_guard.py
javajawa: (My inclination is to move it to `v1` because we might be able to have it applied automatically, so users can just declare their functions and stuff _works_)

ajCameron: That makes sense to me - I'll move it into v1

src/mewbot/api/type_guard.py
ajCameron: ....

It was, as it turns out, not.

I have, however, hacked it. With _moderate_ python crimes.

tests/api/test_type_guard.py
ajCameron: Will do! (Now I have a working form of the code at all).

src/mewbot/api/type_guard.py
ajCameron: Naming is hard...I'll think some more on it.

src/mewbot/api/type_guard.py
ajCameron: `pre_filter_non_matching_events` sounds good to me! I'll make the change.

ajCameron:
I ... am kinda amazed that worked on all versions....

tests/api/test_type_guard.py
ajCameron: `from __future__ import annotations` does not seem to allow it in 3.9 - I'm gonna gate it behind a version check.

ajCameron:
...

....

....

_huh_.

exachixkitsune (commented):
This all looks a bit like magic to me - I'll delve into more detail at some point but don't let me stop you merging it.

javajawa:
@ajCameron I've made some changes -- mostly I've made the two test groups identical except for the `from __future__ import annotations`, and increased the number of tested combinations. If you're happy with what I've done, I'll rebase away most of the commits (I don't think the history is relevant outside of this PR, especially with #206 being raised) and then merge

javajawa:
I have updated and squashed the merge, with a lot of simplification now we don't require python 3.9 support. However, I have not checked that all the test cases are accurately named, or have correct comments.

ajCameron:
I'm very happy with the direction - I'll aim to have a more thorough look over it tomorrow or Sunday.

tests/api/type_guard/test_type_guard_invalid_config.py
```diff
@@ -0,0 +1,160 @@
...
+                return not event
+
+    def test_malformed_trigger_bad_event_type_trigger_annotations(self) -> None:
+        """
+        Tests applying the decorator to a function without "event" as a possibility.
+
+        :return:
+        """
+        with pytest.raises(
+            TypeError, match='can only concatenate str \\(not "int"\\) to str'
```
javajawa:
```suggestion
            TypeError, match=r'(can only concatenate str \\(not "int"\\) to str|Forward references must evaluate to types. Got 7.)'
```

tests/api/type_guard/test_type_guard_invalid_config.py
```diff
@@ -0,0 +1,160 @@
...
+                return not event
+
+    def test_malformed_trigger_bad_event_type_trigger_annotations(self) -> None:
+        """
+        Tests applying the decorator to a function without "event" as a possibility.
+
+        :return:
+        """
+        with pytest.raises(
+            TypeError, match=r'(can only concatenate str \\(not "int"\\) to str|Forward references must evaluate to types. Got 7.)'
```
javajawa:
```suggestion
            TypeError, match=r'(can only concatenate str \(not "int"\) to str|Forward references must evaluate to types. Got 7.)'
```

javajawa (approved):

PR #205, opened 2023-06-16
Enhanced plugin packaging - adding setup.py e.t.c. to plugins - adding discord-dice-roller example

ajCameron:
- added functional setup.py files for all the plugins
 - determine installation order to achieve a functioning venv using the setup.py scripts for mewbot and all the plugins
 - constructed a helper script to build and test such a venv
 - (helper script is deliberately stand alone - with no dependencies outside the standard library or from elsewhere in mewbot)
 - more work can be done - but the helper script and plugin setup.py files are feature complete

Submitting this to check direction before this continues to grow as a project.

### DIFF
--- a/plugins/mewbot-discord_dice_roller/requirements.txt
+++ b/plugins/mewbot-discord_dice_roller/requirements.txt
@@ -2,5 +2,5 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-mewbot.api
+mewbot-api
 d20
--- a/dev-docs/plugin-dev-notes/README.md
+++ b/dev-docs/plugin-dev-notes/README.md
@@ -21,7 +21,7 @@ It's designed to take yaml files, load them, and turn them into bots which can i
 While it does have some basic capabilities, the majority of its capabilities come from plugins
 (either plugins already bundled with the program or plugins written for it.)
 
-If you want to extend the capabilities of the _framework itself_ you want to write a **mewbot framework plugin**.
+Thus, if you want to extend the capabilities of the _framework itself_ you probably want to write a **mewbot framework plugin**.
 
 On a technical level this will take the form of a namespace plugin which will use the `mewbot_framework_plugin_template`.
 
@@ -29,6 +29,10 @@ E.g. I would like mewbot to be able to talk to "Insert communications protocol h
 I will write a plugin which will present a new IOConfig in `mewbot.io`
 
 For further information on how to write namespace plugins, please see the python core documentation - [python namespace plugin docs](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/).
+
+#### 2) Creating tools for bots
+
+If, on the other hand, you want to write some tools which bots _themselves_ will use, then you probably want to write a **mewbot bot plugin**.
 
 Your plugin will be used by installing your module and then calling the IOConfig using something like
 
@@ -41,10 +45,6 @@ properties:
 
 ---
 ```
-
-#### 2) Creating tools for bots
-
-If, on the other hand, you want to write some tools which bots _themselves_ will use, then you may want to write a **mewbot bot plugin**.
 
 This would be a plugin for mewbot which does not add to the core module namespace.
 It might include detailed Triggers, Conditions and Actions that you will use to build your bot via yaml.
@@ -70,6 +70,7 @@ In fact, please do!
 That would really be very helpful.
 
 Your generic IOConfig will take the form of a namespace plugin.
+
 It will be used by importing it from the wider mewbot namespace via something like this
 
 ```python
@@ -86,6 +87,7 @@ Structure it how you like - it matters a lot less for this sort of plugin than a
 (Though there is an example in plugins under `mewbot-discord_dice_roller` which should serve to get you started).
 
 Use the components in your bot by having lines such as
+
 
 ```yaml
 kind: Behaviour
@@ -110,7 +112,6 @@ actions:
     properties: { }
 
 ```
-
 in your yaml files.
 
 (descriptive names are preferred for this sort of plugin as the yaml is intended to be highly human readable).
@@ -134,4 +135,6 @@ Which, for every service, vastly increases the number of useful bots than can be
 Bot components allow the bots to do more things.
 They provide capabilities, and allow each bot written for mewbot to do more.
 
-Both are very helpful!
+Both are very useful!
+
+If you have any issues or doubts, feel free to contact us and we'll try to help!
--- a/plugins/mewbot-discord_dice_roller/LICENSE.md
+++ b/plugins/mewbot-discord_dice_roller/LICENSE.md
@@ -1,0 +1,1 @@
+../../LICENSES/BSD-2-Clause.txt
--- a/plugins/mewbot-io-discord/LICENSE.md
+++ b/plugins/mewbot-io-discord/LICENSE.md
@@ -1,0 +1,1 @@
+../../LICENSES/BSD-2-Clause.txt
--- a/plugins/mewbot-io-discord/README.md
+++ b/plugins/mewbot-io-discord/README.md
@@ -1,0 +1,23 @@
+<!--
+SPDX-FileCopyrightText: 2023 Mewbot Developers <mewbot@quicksilver.london>
+
+SPDX-License-Identifier: BSD-2-Clause
+-->
+
+# mewbot-io-discord
+
+This module serves two purposes
+
+1) adds discord functionality to mewbot
+
+2) serves as an example of how to configure a mewbot namespace plugin
+
+A mewbot namespace plugin is a plugin used to extend the framework itself.
+Mostly in the context of additional IOConfigs.
+After installing this module, you'll be able to import the discord IOConfig directly from mewbot.io as follows
+
+```python
+from mewbot.io.discord import DiscordIO
+```
+
+This is - typically - how new IOConfigs - interfaces to allow mewbot to talk to more protocols - should be written.
--- a/plugins/mewbot-io-discord/requirements.txt
+++ b/plugins/mewbot-io-discord/requirements.txt
@@ -4,3 +4,6 @@
 
 # Discord Module
 py-cord~=2.4.0
+
+# Mewbot
+mewbot-api
--- a/plugins/mewbot-io-discord/setup.py
+++ b/plugins/mewbot-io-discord/setup.py
@@ -4,29 +4,20 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-"""
-Installs the mewbot-discord_dice_roller example package on your system.
-
-Note - when building a development environment, use of this file can be a little annoying.
-In particular, it generates an egg-info folder in the same root as this file.
-Which confuses some linters (in particular, mypy).
-It's recommended that, if your preparing a mewbot dev environment, use the script
-src/mewbot/tools/dev_venv.py
-"""
-
 import os
 from pathlib import Path
 
 import setuptools  # type: ignore
 
-# Finding the right README.md
-plugin_dir = Path(__file__).parent
-root_repo_dir = plugin_dir.parent.parent
+# Finding the right README.md and inheriting the mewbot licence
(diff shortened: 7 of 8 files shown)


.gitignore
```diff
@@ -31,3 +31,5 @@ examples/TrivialDiscordBot/secrets.json
 /reports/
 /.pytest_cache/
 /junit.xml
+3.8.0
+4
```
javajawa: What are these? Are they likely to occur to other people?

plugins/mewbot-discord_dice_roller/requirements.txt
```diff
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: 2023 Mewbot Developers <mewbot@quicksilver.london>
+#
+# SPDX-License-Identifier: BSD-2-Clause
 
-mewbot.api
+mewbot-api==0.0.1
```
javajawa: This feels risky in the long term.
When we sort out the versioning question, and version numbers start going up, this might result in a mismatch to the local dev version.

plugins/mewbot-discord_dice_roller/setup.py
```diff
@@ -11,65 +11,61 @@
...
+assert root_repo_dir.exists()
+
+with open(current_file.parent.joinpath("README.md"), "r", encoding="utf-8") as rmf:
+    long_description = rmf.read()
+
+with open(root_repo_dir.joinpath("LICENSE.md"), "r", encoding="utf-8") as lf:
+    true_licence = lf.read()
+
+with open(current_file.parent.joinpath("requirements.txt"), "r", encoding="utf-8") as rf:
+    requirements = list(x for x in rf.read().splitlines(False) if x and not x.startswith("#"))
```
javajawa: I'd extract `current_file.parent` to `plugin_dir`.
```suggestion
# Finding the right README.md and inheriting the mewbot licence
plugin_dir = Path(__file__).parent
root_repo_dir = plugin_dir.parent.parent

assert root_repo_dir.exists()

with (plugin_dir / "README.md").open("r", encoding="utf-8") as rmf:
    long_description = rmf.read()

with (root_repo_dir / "LICENSE.md").open("r", encoding="utf-8") as lf:
    true_licence = lf.read()

with (plugin_dir / "requirements.txt").open("r", encoding="utf-8") as rf:
    requirements = list(x for x in rf.read().splitlines(False) if x and not x.startswith("#"))
```

plugins/mewbot-discord_dice_roller/setup.py
```diff
@@ -11,65 +11,61 @@
...
+# Finding the right README.md and inheriting the mewbot licence
+current_file = Path(__file__)
+root_repo_dir = current_file.parents[2]
+assert root_repo_dir.exists()
+
+with open(current_file.parent.joinpath("README.md"), "r", encoding="utf-8") as rmf:
+    long_description = rmf.read()
+
+with open(root_repo_dir.joinpath("LICENSE.md"), "r", encoding="utf-8") as lf:
+    true_licence = lf.read()
```
javajawa: Hmmm, license handling is interesting. In generally in-tree plugins are likely to be the same license, but I think it would be most correct to symlink back to `./licenses/[IDENT].md` in each folder. That also then means that the code doesn't change when this moves out of the repo (because it doesn't need to know about the rest of the repo).

(Also, one neat thing about the symlink, on platforms that support it -- it allows us to automatically dervive the license identifier -- see next comment)

plugins/mewbot-discord_dice_roller/setup.py
```diff
@@ -11,65 +11,61 @@
...
+    description="Example discord dice roller for mewbot",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/mewbotorg/mewbot",
+    project_urls={
+        "Bug Tracker": "https://github.com/mewbotorg/mewbot/issues",
+    },
+    classifiers=[
+        "Programming Language :: Python :: 3",
+        f"License :: OSI Approved :: {true_licence}",
```
javajawa: This currently contains the entire license, not the identifier.

plugins/mewbot-discord_dice_roller/setup.py
```diff
@@ -11,65 +11,61 @@
 It _should_ also work for local installation, but this functionality is not tested.
```
javajawa: This comment (and the few lines above) is nominally out of date as of this MR (because you've made it workable locally)

src/mewbot/tools/dev_venv.py
```diff
@@ -0,0 +1,632 @@
...
+    try:
+        target_dir: str = sys.argv[1]
+    except IndexError as exp:
+        raise NotImplementedError(
+            "You need to provide a path to a location to build the venv in"
+        ) from exp
+
+    target_dir = os.path.abspath(target_dir)
+    if not os.path.exists(target_dir):
+        raise NotImplementedError(f"{target_dir = } must exist.")
```
javajawa: These errors are...not especially readable? (I'd honestly throw `argparse` at the situation.)

```
$ python src/mewbot/tools/dev_venv.py 
Using MEWBOT_REPO_PATH = '/home/benedict/src/javajawa/mewbot'
With TOOLS_PATH = '/home/benedict/src/javajawa/mewbot/tools'
==== 0 - Validating Input ======================================================
Traceback (most recent call last):
  File "/home/benedict/src/javajawa/mewbot/src/mewbot/tools/dev_venv.py", line 224, in _validate_input
    target_dir: str = sys.argv[1]
                      ~~~~~~~~^^^
IndexError: list index out of range

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/benedict/src/javajawa/mewbot/src/mewbot/tools/dev_venv.py", line 632, in <module>
    sys.exit(0 if main() else 1)
                  ^^^^^^
  File "/home/benedict/src/javajawa/mewbot/src/mewbot/tools/dev_venv.py", line 137, in main
    venv_path = _validate_input()
                ^^^^^^^^^^^^^^^^^
  File "/home/benedict/src/javajawa/mewbot/src/mewbot/tools/dev_venv.py", line 226, in _validate_input
    raise NotImplementedError(
NotImplementedError: You need to provide a path to a location to build the venv in
```

src/mewbot/tools/dev_venv.py
```diff
@@ -0,0 +1,632 @@
...
+
+    venv_name = sys.argv[2]
+    assert (
+        re.match(r"^[a-zA-Z0-9_]+$", venv_name) is not None
```
javajawa: hmmm.
https://pypi.org/project/pathvalidate/#filename-filepath-validator-for-argparse does a "better" check. My first instinct for testing of `venv-test` didn't match this, which I didn't expect. And then I realised I don't know what the best regexp would be.

src/mewbot/tools/dev_venv.py
```diff
@@ -0,0 +1,632 @@
...
+
+    subprocess.run(
+        [
+            activation_path,
+        ],
+        shell=True,
+        check=True,
+    )
+
+    return activation_path
```
javajawa: Fun fact: this does not work on Linux -- because activate is meant to be sourced _into_ the current shell.
Specifically it fails because the file is not executable.

Given that it only needs to set a couple of environment variables, might be easier to just mess with `os.environ`

javajawa: I have mananged to make it _runnable_ but not _functional_ -- nothing ever gets installed.
[fixes.patch.txt](https://github.com/mewbotorg/mewbot/files/11778847/fixes.patch.txt)


```
Using MEWBOT_REPO_PATH = '/home/benedict/src/javajawa/mewbot'
With TOOLS_PATH = '/home/benedict/src/javajawa/mewbot/tools'
==== 0 - Validating Input ======================================================
Checking /home/benedict/src/javajawa/mewbot/src and /home/benedict/src/javajawa/mewbot/plugins for egg-info folders.

==== 1 - Building Venv =========================================================
Building venv using current_exec = '/home/benedict/src/javajawa/mewbot/venv/bin/python' in venv_path = '/home/benedict/src/javajawa/mewbot/venv_branch'

==== 2 - Validate Venv =========================================================
venv created - attempt to activate using '. /home/benedict/src/javajawa/mewbot/venv_branch/bin/activate'

==== 3 - Update venv pip =======================================================

==== 4 - Install all deps into the venv ========================================
==================================================
Pip list in current state
==================================================

==== 5 - install the plugins ===================================================
About to attempt to install plugin_name = 'mewbot-io-discord'
About to attempt to install plugin_name = 'mewbot-discord_dice_roller'
About to attempt to install plugin_name = 'mewbot-io-file-storage'
plugin_setup_py_file = '/home/benedict/src/javajawa/mewbot/plugins/mewbot-io-file-storage/setup.py' did not exist - skipping.
==================================================
Pip list in current state
==================================================

==== 6 - Wipe pypi placeholder installs ========================================

==== 7 - Install the sub-modules in editable form ==============================
==================================================

==== 8 - Rehome egg-info folders ===============================================
plugin_src_path = '/home/benedict/src/javajawa/mewbot/plugins/mewbot-io-file-storage/src' in plugin_folder = 'mewbot-io-file-storage' did not exist.

==== 9 - Run tests on the venv =================================================



venv prepared - activate via ". /home/benedict/src/javajawa/mewbot/venv_branch/bin/activate"
```

javajawa (dismissed):
Like a lot of the ideas, doesn't work under linux.

Recommend reworking the script to use `argparse` and activate the venv by setting the appropriate environment variables.

.gitignore
ajCameron: ... I have no clue, and deleting them didn't seem to cause any harm

plugins/mewbot-discord_dice_roller/setup.py
ajCameron: Agreed and adopted.

plugins/mewbot-discord_dice_roller/setup.py
ajCameron: Updated - with advice to use the dev_venv.py script

src/mewbot/tools/dev_venv.py
ajCameron: I was trying to avoid it - had some problems with it and unittest in the past - but I've switched to it and resolved them.

plugins/mewbot-discord_dice_roller/requirements.txt
ajCameron: Hmmm. Yes. That's clearly a problem. I _think_ dropping the version number is the best solution for the moment.

(It doesn't really matter for the venv build script - it'll be wiping over that version with the local one anyways. The packages installed here are just placeholders).

(Don't think it's the right solution here, but pur is kinda cool - https://pypi.org/project/pur/)

plugins/mewbot-discord_dice_roller/setup.py
ajCameron: Switched over to using symlinks for reading the liscence.

Will error _unless_ it's a symlink, but I think that's what we want people to do.

plugins/mewbot-discord_dice_roller/setup.py
ajCameron: Switched over to using symlinks for reading the liscence.

Will error _unless_ it's a symlink, but I think that's what we want people to do.

ajCameron:
Breaking at different points in different linuxs. Needs a bit more attention.

src/mewbot/tools/dev_venv.py
ajCameron: Ported in enough of their code to do a better validation. Credited and licensed under MIT.

src/mewbot/tools/dev_venv.py
ajCameron: ... mildly embaressing that.

Fixed as suggested by "activating" the venv via variable manip.

ajCameron:
Have remove the dev_venv builder script to it's own PR (https://github.com/mewbotorg/mewbot/pull/226).

javajawa (dismissed):

javajawa (approved):

PR #206, opened 2023-06-17
Drop support for python 3.9

javajawa:
- Gains access to [PEP634](https://peps.python.org/pep-0634/) and the `match` statement
- Gains access to [PEP647](https://peps.python.org/pep-0647/) `TypeGuard` types
- Gains access to [PEP604](https://peps.python.org/pep-0604/) Generic union types `X | Y`

More information at https://docs.python.org/3/whatsnew/3.10.html

The primary motivation for this change is it will significantly simplify #203


ajCameron (approved):
It probably is time...

PR #211, opened 2023-06-22
Added isort and completed first run

ajCameron:
- added isort to requirements-dev.txt
 - added defaults to pyproject.toml
 - run isort for the first time - to some considerable changes
 - considerable enough it's probably worth pull requesting
 - as I suspect this will result in rather a lot of conflicts
 - this commit includes a deliberate mistake - to check diff is same format as black and annotation works

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,3 +40,11 @@ fail_under=40.0
 [tool.coverage.html]
 
 show_contexts=true
+
+[tool.isort]
+profile = "black"
+py_version=310
+sections = ['FUTURE', 'TYPING', 'STDLIB', 'THIRDPARTY', 'FIRSTPARTY', 'LOCALFOLDER']
+known_typing = ["typing", "types", "collections.abc"]
+# 9 is black compatible, bu does not look great
+multi_line_output = 3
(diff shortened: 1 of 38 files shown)


pyproject.toml
```diff
@@ -40,3 +40,11 @@ fail_under=40.0
 [tool.coverage.html]
 
 show_contexts=true
+
+[tool.isort]
+profile = "black"
+py_version=310
+sections = ['FUTURE', 'STDLIB', 'TYPING', 'THIRDPARTY', 'FIRSTPARTY', 'LOCALFOLDER']
+known_typing = ["typing", "types"]
```
javajawa: Not saying we have to, but I suspect code I wrote will have a much more minimal diff with
```suggestion
sections = ['FUTURE', 'TYPING', 'STDLIB', 'THIRDPARTY', 'FIRSTPARTY', 'LOCALFOLDER']
known_typing = ["typing", "types", "collections.abc"]
```

(I'll try this in another branch and see what it looks like)

ajCameron: Adopted and run (looks like a reasonable ordering to me).

javajawa (approved):
Love it, squashed it, will merge.

PR #214, opened 2023-06-26
Added bandit security analysis

ajCameron:
- bandit is a security focussed static analysis tool
 - adding it as a stand alone command
 - and part of the preflight toolchain
 - and as a requirement for dev
 - currently disabled low level security warnings - 19 errors found on this level
 - they can be worked out over time


ajCameron:
Does not seem to be a permission issue... will investigate

PR #230, opened 2023-08-13
Hardened linting against non utf-8 characters in the output

ajCameron:
- provides a fallback if a non-utf character is encountered while processing linting output.

### DIFF
--- a/src/mewbot/tools/lint.py
+++ b/src/mewbot/tools/lint.py
@@ -167,7 +167,12 @@ class LintToolchain(BanditMixin):
 
         result = self.run_tool("PyLint", "pylint")
 
-        for line in result.stdout.decode("utf-8").split("\n"):
+        try:
+            result_lines = result.stdout.decode("utf-8")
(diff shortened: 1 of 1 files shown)


src/mewbot/tools/lint.py
```diff
@@ -167,7 +167,12 @@ def lint_pylint(self) -> Iterable[Annotation]:
 
         result = self.run_tool("PyLint", "pylint")
 
-        for line in result.stdout.decode("utf-8").split("\n"):
+        try:
+            result_lines = result.stdout.decode("utf-8")
+        except UnicodeError:
+            result_lines = result.stdout.decode("utf-8", errors="replace")
```
javajawa: Is there a semantic difference to having the try catch as opposed to seeing this by default?

ajCameron: It _did_, before I spent half an hour trying different error handling methods, and decided replace was best.... corrected.

javajawa (approved):

PR #233, opened 2023-08-20
examples: Add a stdio IOConfig for examples

javajawa:
This makes writing examples easier, as users can trigger events by typing into the console, and the output events can be automatically returned also to the console, removing the need for network connectivity and API access tokens for services like Discord


ajCameron (approved):

ajCameron:
'ruff <path>' needs to be changed to 'ruff check <path>'.

PR #250, opened 2023-10-01
examples: Adding examples command line chooser

ajCameron:
- formatting is currently rudimentary
 - two options are supported
    - running an example directly from the command line using the examples module
    - choosing an example to run from a generated list


ajCameron:
Superceded by https://github.com/mewbotorg/mewbot/pull/254

PR #253, opened 2023-10-06
tools: Dropping mewbot.tools

ajCameron:
Restructuring to use mewbot-dev-tools in place of mewbot.tools.
 - Same code, more generic and useful package.
 - Adding the ruff cache to git ignore
 - Stripped out the tools folder and mewbot.tools entirely


ajCameron:
Turns out cannot use our sonarcloud credentials in this sort of pr...

PR #254, opened 2023-10-06
tools: Removing tools

ajCameron:
Switching over to using mewbot-dev-tools


.github/workflows/review.yaml
```diff
@@ -50,13 +50,65 @@ jobs:
...
 
+      - name: Install mewbot-dev-tools
+        run: python -m pip install mewbot-dev-tools
+        shell: sh
+
+      - name: Install dependencies into default python
+        shell: bash
+        run: python -m mewbot_dev_tools.install_deps
+
+      - name: Run Tests (without install)
```
javajawa: This job feels mislabelled, as the deps would be installed above?

(Which leads to a question, but might be "not yet finished": what's going on with this section in general? Seems overly complicated?
Looking at the dev tools package, `mewbot_test` should be available as an executable...but I guess we still have to get the path stuff (but we can always do that as a bootstrap task in the tools?))

ajCameron: - labelling improved
 - matrices have been simplified
 - it's more complicated than I'd like, but I thought splitting the tasks down into seperate lint and test matrices made it more readable
 - I considered doing the path hacking in tools, but it seemed better to do it explicitely in the action
 - Ideally we'd run the tests both ways - after getting mewbot on the path via path hacks and after running setup.py
 - But I wanted to discuss that change before implementing it - so made a minimal set of changes to the workflow - just enough to get it running with the new tools package

javajawa: I've taken a stab at recombining the two in c5f51d1.
The workflow works, but I've noticed a couple of...oddities (that are likely going to need fixing on the dev tools side):

 - Coverage report does not include `src/mewbot/io/desktop_notification.py`. Might be that there's no tests, but I would expect it to have a 0% entry in the table ([ref](https://github.com/mewbotorg/mewbot/actions/runs/6453497714/job/17517174505#step:6:177))
 - non-zero exit code of `istort` isn't being picked up ([ref 1](https://github.com/mewbotorg/mewbot/actions/runs/6453497714/job/17517173968#step:6:32) | [ref 2](https://github.com/mewbotorg/mewbot/actions/runs/6453497714/job/17517173968#step:6:76))
 - At some point, we should tidy up that `arg_list` output in the dev tools

.github/workflows/review.yaml
```diff
@@ -89,10 +100,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Setup Python ${{ matrix.version }}
+      - name: Setup Python ${{ env.DEFAULT_PYTHON }}
```
javajawa, line 101: Fun thing: because the annotations output is really simple code, having a random python version was only breaking the pip cache. Which I noticed because this stage was also installing all the mewbot dependencies for a bit and that takes _forever_.

requirements-dev.txt
```diff
@@ -10,29 +10,20 @@
...
 pydeps~=1.12.1
 myst-parser~=2.0.0
 sphinx~=7.2.6
 m2r2~=0.3.3.post2
 
 # Python Linting toolchain
-black>=23.1.0
-flake8>=6.0.0
-pylint>=2.16.0
 pytest-github-actions-annotate-failures>=0.1.7
```
javajawa, line 14: Is this still used? Or does our annotations stuff take care of this exclusively?

.github/workflows/review.yaml
ajCameron: Ahhhhh! Good point.

ajCameron: It does indeed! Good change.

requirements-dev.txt
ajCameron: It should be taken care of either there or in tools.

Some more deps should not be required. I'll pare the file back.

ajCameron: Pare back complete

.github/workflows/review.yaml
ajCameron: Indeed - this'll have to be solved on the mewbot-dev-tools side.

Will fix 2 and 3.

Point 1 seems to boil down to pytest not handling namespace plugins correctly. Will experiment and find a workaround (as I suspect this will happen with any custom namespace package as well ... which is suboptimal)

ajCameron: Point 1 haos been traced to a - seeming - issue in how coverage itself handles namespace plugins (it doesn't - not correctly, anyhow).

I have a extremely hacky solution
 - on run time, introspect to generate a new test file in the test tree
 - said test file just contains imports for all the namespace plugins found in mewbot.io
 - this seems to work - but might be a cure worse than the disease

ajCameron: 2 and 3 have been addressed in mewbot-dev-tools 0.0.13

javajawa (approved):

PR #258, opened 2023-10-15
main_module: Adding __init__.py and version bump

ajCameron:


ajCameron:
On further investigation, turned out to be a weirdness in how windows handles paths.

So it goes.

